PR #6, opened 2021-02-03
Bump rails version requirement for compact_blank

itsliamegan:
ActiveSupport's `Enumerable#compact_blank` (used [here](https://github.com/rails/kredis/blob/a7e64f3d0efecae906f78a836bb6a10ea1183000/lib/kredis/types/proxy.rb#L22)) wasn't added until [this commit](https://github.com/rails/rails/commit/c8847c17a7c9ae75c44c522c56ccd9c5fca25ea7), which wasn't included until Rails 6.1 as per the [changelog](https://github.com/rails/rails/blob/v6.1.1/activesupport/CHANGELOG.md). I bumped the requirement to Rails 6.1 in the gemspec in order to account for this. I understand if you want to support Rails 6.0, in which case I can instead just change `compact_blank` to `reject(&:blank?)`

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     kredis (0.2.0)
-      rails (>= 6.0.0)
+      rails (>= 6.1.0)
       redis (~> 4.0)
 
 GEM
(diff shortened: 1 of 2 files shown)


itsliamegan:
Closed as duplicate of #7

PR #7, opened 2021-02-03
compact_blank wasn't added until Rails 6.1

leastbad:
This PR seeks to allow Rails 6.0 users to enjoy Kredis.

### DIFF
--- a/lib/kredis/types/proxy.rb
+++ b/lib/kredis/types/proxy.rb
@@ -18,7 +18,7 @@ class Kredis::Types::Proxy
   private
     def log_message(method, *args, **kwargs)
       args      = args.flatten.reject(&:blank?).presence
-      kwargs    = kwargs.reject(&:blank?).presence
+      kwargs    = kwargs.reject { |_k, v| v.blank? }.presence
       type_name = self.class.name.split("::").last
 
       "[Kredis #{type_name}] #{method.upcase} #{key} #{args&.inspect} #{kwargs&.inspect}".chomp
--- a/lib/kredis/types/proxy.rb
+++ b/lib/kredis/types/proxy.rb
@@ -1,5 +1,3 @@
-require "active_support/core_ext/enumerable"
-
 class Kredis::Types::Proxy
   attr_accessor :redis, :key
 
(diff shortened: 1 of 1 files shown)


lib/kredis/types/proxy.rb
```diff
@@ -19,8 +17,8 @@ def method_missing(method, *args, **kwargs)
 
   private
     def log_message(method, *args, **kwargs)
-      args      = args.flatten.compact_blank.presence
-      kwargs    = kwargs.compact_blank.presence
+      args      = args.flatten.reject(&:blank?).presence
+      kwargs    = kwargs.reject(&:blank?).presence
```
kaspth, line 21: I'm amazed that our test suite still passes because I think this needs to be `reject { |k, v| v.blank? }` as it's called on a Hash.

kaspth (commented):
Nice!

lib/kredis/types/proxy.rb
leastbad: Honestly, I just cut and pasted the contents of `compact_blank` to make it work. Apparently, hash values can be blank!

leastbad: I am seeing some pretty strange behaviour from your counter type:

```
> counter.increment
Traceback (most recent call last):
        1: from (irb):2
ArgumentError (wrong number of arguments (given 1, expected 0))
```

When I run the test suite, it's telling me: `/home/leastbad/kredis/lib/kredis/types/proxying.rb:7: warning: '*' interpreted as argument prefix`

kaspth: > Honestly, I just cut and pasted the contents of `compact_blank` to make it work. Apparently, hash values can be blank!

Not sure what you mean? There's a distinct `Hash` version: https://github.com/rails/rails/blob/d30dad4bc1039c80648f39c5171b146f8dd8abd6/activesupport/lib/active_support/core_ext/enumerable.rb#L199

kaspth: > I am seeing some pretty strange behaviour from your counter type:

I couldn't reproduce that example with:

```ruby
counter = Kredis.counter "mycounter"
counter.increment
```

Do you have the full backtrace?

> /home/leastbad/kredis/lib/kredis/types/proxying.rb:7: warning: '*' interpreted as argument prefix

Ah I guess `rake` is running the tests with warnings enabled, I disagree with the warning so I'd rather not add the parens 😂

leastbad: You were obviously correct about there being a Hash version, which I'd missed.

Latest commit appears to be working great with Rails 6.0. Thanks for being patient.

kaspth:
Sweet, thanks!

PR #18, opened 2021-02-24
Update README for reading the members of a Kredis.set

danott:
The README example for `Kredis.set` was sending the message `elements` to read the members of a set.  Following that example resulted in a runtime error.

Looking into the implementation of [Kredis::Types::Set][], it appears that `members` is the correct method for reading the members?

[Kredis::Types::Set]: https://github.com/rails/kredis/blob/ab79f255c53bfb5e760c08644222af6fe215f253/lib/kredis/types/set.rb#L6-L8


kaspth:
Nice, thank you! `elements` was an earlier version.

PR #19, opened 2021-02-25
Exploring the idea of expiring scalar values

danott:
Redis' feature for expiring keys has surfaced in [counters][] and in [flags][]. I've found myself wanting (needing?) expiring keys for scalars and lists. My consumer-side solution has been shoe-horning my feature into using expiring flags, or writing manual expiration code that reaches into the implementation of `Kredis`.

These commits tell the story of searching for an API to expire scalars in three ways:

1. Expiring a scalar in a relative amount of time by calling `scalar.expire_in 1.day`
2. Expiring a scalar at a specific point in time by calling `scalar.expire_at 2.days.from_now`
3. Configuring a scalar to always expire after assignment with `Kredis.scalar "ephemeral", lifespan: 2.days`

As previously mentioned, I've found myself reaching for this behavior for both scalars and lists so far. I could imagine bringing this API to several (all?) the types. But before getting too invested in the idea, I wanted to share the exploration as a proof of concept, and catalyst for conversation.

Is this a feature `Kredis` would be open to?

[counters]: https://github.com/rails/kredis/blob/5d576221a9aadc6c73aa67229864bce623ac3d54/test/types/counter_test.rb#L39
[flags]: https://github.com/rails/kredis/blob/1cb2f295fdf622cf44e6d906b28a2c61bb7ab326/test/types/flag_test.rb#L16


hayesr:
Came here to raise the same question and found this. I'd like to use Kredis lists to power a simple background job dependency system. These lists would only need to live for about a day. We ran into a timeout problem trying to clean up old keys. If they could expire themselves it would be an easy solution.

dhh:
I'd be happy to see this, but we should match the API to follow what we already have. So that means adding expire_in/expire_at at the definition phase, rather than introduce "lifecycle" as a new term.

dhh:
As part of documenting this, could you give some examples of how you've used expiring scalars in your apps?

dhh:
Just coming back to this. I still dig it. Would you mind updating for main? Then I'll get this merged.

lib/kredis/types.rb
```diff
@@ -5,40 +5,38 @@ def proxy(key, config: :shared, after_change: nil)
...
-    type_from(Scalar, config, key, after_change: after_change, typed: :datetime, default: default)
+  def datetime(key, default: nil, config: :shared, after_change: nil, expires_in: nil)
+    type_from(Scalar, config, key, after_change: after_change, typed: :datetime, default: default, expires_in: expires_in)
   end
 
-  def json(key, default: nil, config: :shared, after_change: nil)
-    type_from(Scalar, config, key, after_change: after_change, typed: :json, default: default)
+  def json(key, default: nil, config: :shared, after_change: nil, expires_in: nil)
+    type_from(Scalar, config, key, after_change: after_change, typed: :json, default: default, expires_in: expires_in)
   end
```
danott, line 39: Restore a newline that was removed during merge conflict resolution?

```suggestion


```

lib/kredis/types.rb
```diff
@@ -5,37 +5,36 @@ def proxy(key, config: :shared, after_change: nil)
     type_from(Proxy, config, key, after_change: after_change)
   end
 
-
-  def scalar(key, typed: :string, default: nil, config: :shared, after_change: nil)
-    type_from(Scalar, config, key, after_change: after_change, typed: typed, default: default)
+  def scalar(key, typed: :string, default: nil, config: :shared, after_change: nil, expires_in: nil)
```
danott, line 8: Restore another newline that was removed in the merge conflict resolution.

```suggestion

    def scalar(key, typed: :string, default: nil, config: :shared, after_change: nil, expires_in: nil)
```

PR #20, opened 2021-02-26
Add kredis_boolean attribute with default argument

MrHubble:
Add `kredis_boolean`. Follows in the same vein of existing attributes (e.g. `kredis_integer`). The only difference being I added `default` as an argument:

```
kredis_boolean :noodling, default: false
```

### DIFF
--- a/lib/kredis/attributes.rb
+++ b/lib/kredis/attributes.rb
@@ -14,6 +14,10 @@ module Kredis::Attributes
       kredis_connection_with __method__, name, key, config: config
     end
 
+    def kredis_boolean(name, key: nil, default: nil, config: :shared)
+      kredis_connection_with __method__, name, key, default: default, config: config
+    end
+
     def kredis_datetime(name, key: nil, config: :shared)
       kredis_connection_with __method__, name, key, config: config
     end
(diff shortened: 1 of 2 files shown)


kaspth:
What's the use case? This is generally what the flag type is meant to support, so the boolean aspect is whether or not the key has a value in Redis.

MrHubble:
I didn't realise that flag was meant for this use case.

I'm happy to close this PR if it adds no value. 

Thanks for the feedback. I'm a big fan of your work.

azyzio:
Hi,

Thanks for this great encapsulation :) Had to build a custom framework for this kind of things before.

I can actually see a use case for booleans with defaults:
- we can have several user settings to enable / disable certain features, e.g. `reports_enabled`, `api_enabled`, `notifications_enabled`
- For consistency, we want to keep all naming to have `_enabled` suffix
- By default, we want to keep certain features enabled and others disabled

Right now, we could do it by raising flags in an `after_craete` callback, but a default option would be nicer.

Also, flags seem to be great for 'flagging' things. E.g. `crossed_the_finish_line`, `invitation_sent`. It feels that settings like above are a kind of different category. But this is a minot thing :)

Maybe you can see a better place for settings like those?

Best,
Adam

lastthyme (approved):

PR #21, opened 2021-03-01
add more redis commands to the readme. Fix some incorrect commands

lsylvester:
Adds documentation for the redis commands being executed to more of the examples in the README.

Updates the documented commands where is did not match the logged commands from running in the console.

Fixes the order of the elements  in the final value for the unique list to match the observed behaviour (LPUSH appears to reverse the order of the elements as it pushes them onto the list).

Fixes the documented key name when using attributes to use the plural of the model name, not the singular form.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,21 +29,21 @@ There are data structures for counters, enums, flags, lists, unique lists, sets,
 
 ```ruby
 list = Kredis.list "mylist"
-list << "hello world!"
-[ "hello world!" ] == list.elements
+list << "hello world!"               # => RPUSH mylist "hello world!"
+[ "hello world!" ] == list.elements  # => LRANGE mylist 0, -1
 
 integer_list = Kredis.list "myintegerlist", typed: :integer
-integer_list.append([ 1, 2, 3 ])        # => LPUSH myintegerlist "1" "2" "3"
-integer_list << 4                       # => LPUSH myintegerlist "4"
-[ 1, 2, 3, 4 ] == integer_list.elements # LRANGE 0 -1
+integer_list.append([ 1, 2, 3 ])        # => RPUSH myintegerlist "1" "2" "3"
+integer_list << 4                       # => RPUSH myintegerlist "4"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,21 +29,21 @@ There are data structures for counters, enums, flags, lists, unique lists, sets,
...
-[ 1, 2, 3, 4 ] == integer_list.elements # LRANGE 0 -1
+integer_list.append([ 1, 2, 3 ])        # => RPUSH myintegerlist "1" "2" "3"
+integer_list << 4                       # => RPUSH myintegerlist "4"
+[ 1, 2, 3, 4 ] == integer_list.elements # => LRANGE myintegerlist 0 -1
 
 unique_list = Kredis.unique_list "myuniquelist"
-unique_list.append(%w[ 2 3 4 ])
-unique_list.prepend(%w[ 1 2 3 4 ])
+unique_list.append(%w[ 2 3 4 ])                # => LREM myuniquelist 0, "2" + LREM myuniquelist 0, "3" + LREM myuniquelist 0, "4"  + RPUSH myuniquelist "2", "3", "4"
+unique_list.prepend(%w[ 1 2 3 4 ])             # => LREM myuniquelist 0, "1"  + LREM myuniquelist 0, "2" + LREM myuniquelist 0, "3" + LREM myuniquelist 0, "4"  + LPUSH myuniquelist "1", "2", "3", "4"
```
lsylvester, line 42: As this is doing one command per element + 1 command at the end it might be better to skip some of the commands to make the line shorter like

```
LREM myuniquelist 0, "1"  + LREM myuniquelist 0, "2" + ... + LPUSH myuniquelist "1", "2", "3", "4"
```

if you still think it is clear to the reader.

lsylvester:
Also - the datetime example wasn't working for me because `DateTime.tomorrow` and `DateTime.yesterday` are returning `Date`s instead of `DateTime`s which are not able to be deserialized when read out of redis.

I am not sure if this should be fixed by having `DateTime.tomorrow` return a DateTime, allowing Dates to be serialized in a DateTime redis store - or update the example to not have methods that are returning dates.

PR #29, opened 2021-04-15
Order the Ruby `require` statements and the Gems

jbampton:
Remove unneeded blank lines from Markdown and Ruby.


kaspth:
They are ordered intentionally like this, thanks.

PR #32, opened 2021-05-21
Utilize ActiveSupport instrumentation for logging

npezza93:
Using ActiveSupport instrumentation allows us to track the duration and
have more control over logging.


kaspth:
Awesome, I dig this — nice work!

Issue #17, opened 2021-02-16
Is it possible to add ActiveRecord-like change callbacks?


julianrubisch:
*Disclaimer: I'm not a Redis expert, at all.*

My naive line of thought was that change callbacks could be used in a variety of scenarios:

- issuing notifications, e.g. when a user is added to a unique list holding "presence"
- updating a view via turbo streams or CableReady, think "live dashboards"
- building chat bots
- ...

kaspth:
Hey, yes, there's something interesting here! Right now this gets close to speculative territory though, so let's get a little more real by seeing some use cases from your app. Is there a place where you're trying to add Kredis that you could reasonably share (with renames/modifications)? That way we can extract rather than invent, which is easier 😄

julianrubisch:
Very speculative indeed, because right now I'm only building this in my head - as kredis itself is quite young. I'm going to build a presence module in one of my apps soon-ish though, so I'll report back with a gist when I'm ready 👍

kaspth:
Perfect that sounds great! I'll close this for now, but once you have more we can just re-open, or start a new one, all good 😄

leastbad:
Redis (and the `redis`) gem make it relatively easy to "subscribe" to updates on a key:

https://github.com/redis/redis-rb/blob/master/examples/pubsub.rb

I think you/we get further, faster, if you think of this potential functionality as being less like an AR callback and more like an ActionCable channel.

Right now, if I was 1. working with a Kredis object eg. `Kredis.string "mystring"` I would likely just take the example linked above and use the Redis gem directly to subscribe and respond to updates.

Conversely, if your goal is to 2. get the AR methods like `kredis_list` to trigger AR callbacks on their models, well, that's interesting. You could likely prototype this as a concern and whip up an example pretty quickly, Julian.

(If there's a third path implied, I apologize for not spotting it!)

julianrubisch:
I managed to play around with this a bit more, and it sent me right down a rabbit hole of ThreadPools etc., looking at the ActionCable pubsub implementation - stuff I'm not really too comfortable with.

Also, since kredis is supposed to be a lightweight adapter for redis data types, a complete pubsub solution doesn't appear to have a place here.

What could bring some flexibility though, is the ability to pass a callback block to mutating operations like `Set#add` etc.

A very naive solution which has no concept of thread safety etc. could look like this:

```rb
class Kredis::Types::Set
  # ...
  def add(*members)
    sadd types_to_strings(members) if members.flatten.any?

[...]

PR #33, opened 2021-06-11
Proposal: Yield elements to block for collection mutations

julianrubisch:
This PR adds the changes mentioned in https://github.com/rails/kredis/issues/17#issuecomment-852177492

Doing that I was able to write a simple presence module that streams a CableReady operation to display a list of users currently viewing a resource:

```rb
resource.present_users.add(current_user.id) do |members|
  StreamPresenceJob.perform_later(members)
end
```

From here it's not a wide stretch to assume that a `Proc` could be placed on the model itself, or a wrapper function defined inside a concern for adding/removing ids to/from the set that allows for this operation.

I think this is a great, lightweight way to report the result of a mutation back to the caller. For the record, simply calling `.members` after an `.add` operation didn't cut it, assumingly because of threaded operation.

### DIFF
--- a/lib/kredis/types/list.rb
+++ b/lib/kredis/types/list.rb
@@ -10,14 +10,22 @@ class Kredis::Types::List < Kredis::Types::Proxying
 
   def remove(*elements)
     types_to_strings(elements).each { |element| lrem 0, element }
+
+    yield send(:elements) if block_given?
   end
 
   def prepend(*elements)
     lpush types_to_strings(elements) if elements.flatten.any?
+
+    # if called from a subclass/in a multi, don't yield from here
+    yield send(:elements) if block_given? && !(self.class < Kredis::Types::List)
   end
 
   def append(*elements)
     rpush types_to_strings(elements) if elements.flatten.any?
+
+    # if called from a subclass/in a multi, don't yield from here
+    yield send(:elements) if block_given? && !(self.class < Kredis::Types::List)
   end
   alias << append
 end
--- a/lib/kredis/types/set.rb
+++ b/lib/kredis/types/set.rb
@@ -10,11 +10,15 @@ class Kredis::Types::Set < Kredis::Types::Proxying
 
   def add(*members)
     sadd types_to_strings(members) if members.flatten.any?
+
+    yield send(:members) if block_given?
   end
   alias << add
 
   def remove(*members)
     srem types_to_strings(members) if members.flatten.any?
+
+    yield send(:members) if block_given?
   end
 
   def replace(*members)
@@ -22,6 +26,8 @@ class Kredis::Types::Set < Kredis::Types::Proxying
       del
       add members
     end
+
+    yield send(:members) if block_given?
   end
 
   def include?(member)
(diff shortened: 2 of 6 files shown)


dhh:
I'm not quite sure I follow the intention here. How is this different from first calling #add, then calling #members? I find the yield approach pretty confusing, but perhaps that's why I don't understand the value over just making two method calls?

julianrubisch:
As I commented, I had concurrency issues when I did that - presumably they're executed in different threads? Might be totally my fault/misunderstanding but it sent me searching for a solution. 

Also: the ability to define an "after_add" proc elsewhere and inject it seemed tempting.

True, could wrap it all myself etc, but it seemed easy enough.

If you don't see value in it that's totally fine, I will then just try to craft an abstraction on top of it.

dhh:
Ah, I see what you mean. I think we need to expose something higher-level, then, like redis.multi. But let it be its own structure, rather than a side effect via yield.

julianrubisch:
Ah, like a multi adapter/wrapper?

I'm no Redis expert, would you wrap everything in a `multi` and yield from in there? Because I tried and you get a Redis::Future...

Or rather... return the result of a `multi`?

dhh:
Wrapping everything in multi will ensure that all the commands are executed sequentially, and it'll be faster too. Then you return the result of the last command 👍

julianrubisch:
Gotcha, IIUC you can call `.value` on a Future (result of an operation inside a `multi`)

What would be the preferred API then? I had the idea of including/pretending a module that would expose a class method `returns_elements_for` and would define instance methods

`returns_elements_for :add` => `define_method :returning_add { multi do ... end }`

(Better naming TBD)

dhh:
Maybe something very parallel, like:

```ruby
members = resource.present_users.multi do |set|
  set.add(current_user.id)
  set.members
end

StreamPresenceJob.perform_later members
```

Although, looking at this, we really shouldn't have to do this for concurrency issues. I don't actually understand why you'd get out of order responses. It feels like that's actually the bug we should fix, rather than try to pave over it with UI.

I mean, multi can still make sense from a performance perspective. But the case above it does not seem to warrant such an optimization.

julianrubisch:
Agreed re concurrency, it might easily have been some kind of glitch on my side.

The initial motivation though was to have a lightweight way to respond to changes, see #17

@leastbad pointed out that a PubSub solution would be most robust, and I agree, but it seems out of bounds for this library.

dhh:
Gotcha. I think we're still missing a realistic case of where this is needed/desired. And to contrast that with simply wrapping things in a method call. We've used Kredis a lot, and so far haven't found a case that could be improved with a callback like this.

julianrubisch:
No question there. Syntactic sugar, certainly.

Okay, so let's start over: IMHO the use case is very realistic: I'd like the `Kredis::Set` to inform "observers" on change.

Which is exactly the scenario here: run a job whenever a user has been added to the `present_users` set.

Can I just craft a convenience method on the model that would enable a model to emit those type of events? Absolutely. I just thought it could be a nice addition to the library 🤷‍♂️

dhh:
Gotcha, yeah, I do think there's a basic callback setup here that just uses the Active Model callback structure. Would want to find something high-level and generic that preferably would work with all the types. Ala `changed`.

julianrubisch:
Yes! Maybe it's my own weird attachment to inversion of control, but in my way of thinking the (for example) Kredis::Set knows when it changed and should inform interested parties.

Anyways, no harm done, learned a lot about Redis on the way, too. 👍🏻

PR #34, opened 2021-06-13
Changed Callbacks

julianrubisch:
Following up with #33, here's another take on the API.

The `Set` initializer is passed an (optional) `changed:` argument that can hold a lambda.

@dhh Obviously I'll wrap this in a more DRY abstraction, just wanted to check back if this POC aligns better with your concept of the library?

*Edit:* Uses `ActiveSupport::Callbacks` now ✌️

### DIFF
--- a/lib/kredis/attributes.rb
+++ b/lib/kredis/attributes.rb
@@ -38,16 +38,16 @@ module Kredis::Attributes
       kredis_connection_with __method__, name, key, config: config
     end
 
-    def kredis_list(name, key: nil, typed: :string, config: :shared)
-      kredis_connection_with __method__, name, key, typed: typed, config: config
+    def kredis_list(name, key: nil, typed: :string, config: :shared, changed: ->(list){})
+      kredis_connection_with __method__, name, key, typed: typed, config: config, changed: changed
     end
 
-    def kredis_unique_list(name, limit: nil, key: nil, typed: :string, config: :shared)
-      kredis_connection_with __method__, name, key, limit: limit, typed: typed, config: config
+    def kredis_unique_list(name, limit: nil, key: nil, typed: :string, config: :shared, changed: ->(list){})
+      kredis_connection_with __method__, name, key, limit: limit, typed: typed, config: config, changed: changed
     end
 
-    def kredis_set(name, key: nil, typed: :string, config: :shared)
-      kredis_connection_with __method__, name, key, typed: typed, config: config
+    def kredis_set(name, key: nil, typed: :string, config: :shared, changed: ->(set){})
+      kredis_connection_with __method__, name, key, typed: typed, config: config, changed: changed
     end
 
     def kredis_slot(name, key: nil, config: :shared)
--- a/lib/kredis/types.rb
+++ b/lib/kredis/types.rb
@@ -53,8 +53,8 @@ module Kredis::Types
     Enum.new configured_for(config), namespaced_key(key), values: values, default: default
   end
 
-  def list(key, typed: :string, config: :shared)
-    List.new configured_for(config), namespaced_key(key), typed: typed
+  def list(key, typed: :string, config: :shared, changed: ->(list){})
+    List.new configured_for(config), namespaced_key(key), typed: typed, changed: changed
   end
 
   def unique_list(key, typed: :string, limit: nil, config: :shared)
--- a/lib/kredis/types.rb
+++ b/lib/kredis/types.rb
@@ -57,8 +57,8 @@ module Kredis::Types
     List.new configured_for(config), namespaced_key(key), typed: typed, changed: changed
   end
 
-  def unique_list(key, typed: :string, limit: nil, config: :shared)
-    UniqueList.new configured_for(config), namespaced_key(key), typed: typed, limit: limit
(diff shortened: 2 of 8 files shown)


julianrubisch:
Can confirm that using this I was able to condense my Presence handling into this:

```rb
module UserPresence
  extend ActiveSupport::Concern

  included do
    kredis_set :present_users, changed: ->(users) do
      StreamPresenceJob.perform_later(users: users)
    end
  end
end

class PresenceChannel < ApplicationCable::Channel
  def subscribed
    resource = GlobalID::Locator.locate_signed params[:signed_resource]
    if resource.present?
      stream_from "presence:#{resource.id}"
      resource.present_users.add(current_user.id)
    else
      reject
    end
  end

  def unsubscribed
    resource = GlobalID::Locator.locate_signed params[:signed_resource]
    return unless resource.present?

    resource.present_users.remove(current_user.id)
  end
end
```

which I personally find super slick 🧙

dhh:
Nice! Think we need to level this up so it also takes method references, just like AR does. So you can have `changed: :stream_presence_later`.

julianrubisch:
Exactly my thoughts, ++, I'll tackle that next.

julianrubisch:
Please correct me if I'm wrong, but IIUC this would require a `Kredis::Something` to hold a reference to the model it was created by?

Otherwise it wouldn't be possible to access the context of the model?

It could be obtained by reversing `kredis_key_for_attribute(name)`, but I'm not sure if that is desirable. Else I don't know how to access it...

dhh:
Hmm, yeah, fair point. That might be a step too far. Perhaps just passing the function is fine.

julianrubisch:
I have an update in ~10 minutes :-)

julianrubisch:
I added a class method `runs_callbacks_for` that dynamically defines new methods, and wraps `run_callbacks :change` around `super` (https://github.com/rails/kredis/pull/34/files#diff-a116f0518441926828bcb875bcddbd01cebb0942818e6e3ea2f931cc4d741bd1R5)

It's not done yet because it seems there is an issue with kwargs (`Counter.increment(by:)`, for example), but I think this should come out elegantly.

LMK what you think!

lib/kredis/attributes.rb
```diff
@@ -38,16 +38,16 @@ def kredis_json(name, key: nil, config: :shared)
       kredis_connection_with __method__, name, key, config: config
     end
 
-    def kredis_list(name, key: nil, typed: :string, config: :shared)
-      kredis_connection_with __method__, name, key, typed: typed, config: config
+    def kredis_list(name, key: nil, typed: :string, config: :shared, changed: ->(list){})
```
dhh, line 41: Should just use nil as the default, so we're not creating new enclosures when they're not used.

dhh:
Given the fact that we're only exposing a single callback, I don't think it actually warrants using the Active Model callback setup. We're not using any of the filters either.

I think we should follow the naming convention, though, so it's `kredis_list :users, after_change: ->(users) { }`.

But hmmm.. Looking at that, I think it's pretty iffy that you don't have access to the instance owning the kredis connection. Which means it's really only usable for something like a job invocation. Not sure that offers quite enough benefit.

I totally see where you're coming from, though. You don't want to wrap three mutating methods on a list to be able to do something on any change. So I think the demand is right. But don't think we've hit on the right implementation yet. Because we have to weigh this with the complexity we're introducing.

Another potential path is to wrap the Kredis type in a proxy on the defining object side. Which could then invoke the callback.

julianrubisch:
So you're essentially proposing that this be solved as a kind of kredis plugin?

So you'd essentially wrap the type in an adapter on demand and inject behavior there?

dhh:
I'm thinking that this could be solved by having a proxy wrapping the normal Kredis type object when it's setup by the `kredis_list` class method. Then that proxy can hold a reference to both the defining object and the type, and thus easily be able to invoke instance methods on the defining object.

julianrubisch:
Yep, that's along my initial thoughts that I could implement this as an AR type concern but then I thought that probably don't want Kredis to have any notion of where it's being instantiated from.

Alright let's do this 👏🏻

dhh:
Yeah, part of the balance here is not to let a relatively niche concern infect too much other stuff. So having a single proxy handle all of it, with as little as possible bleeding into everything else, would be good.

julianrubisch:
Very thoughtful. The question then that comes to my mind is: shouldn't that really be another gem?

I know that I brought this up here because I thought the whole lib could benefit from it, but at this point I could really just make it a separate gem and just do some monkey patching 😁

dhh:
I do think there's benefit here! And another gem is imo too much separation. I think we can get there with a proxy that owns all the logic, and just takes the configuration from the definition. There's a line here. It's just relatively thin 😄

julianrubisch:
Yeah agreed.

So let's say we call this a `Kredis::CallbacksProxy`, it holds a wrapped `type` and a `caller` (naming is hard), the operations which should be wrapped would still be defined by the `type`, right, possibly just by a frozen class attr reader?

```rb
class Kredis::Set ...
  def self.callback_operations
    %i[add ...].freeze
  end 
end 
```

Those could be used to wrap the calls, and callback could then be executed by just calling `to_proc` on whatever is passed in - in theory.

PR #36, opened 2021-07-08
Provide factory method for vanilla kredis types

julianrubisch:
Building up on #35, I just wanted to quickly PR the factory method for vanilla types as mentioned in the PR comments, to address @leastbad's and others' concerns.

The only meaningful changes are:
1. a new factory method `Kredis.with_callback` that you can pass a type as a symbol, plus key, arguments, and callback,
2. the proc is now called with `@record || @type`, making sure that if not being instantiated from an ActiveRecord object, it's passing the kredis type back to the lambda as an argument.

### DIFF
--- a/lib/kredis/callbacks_proxy.rb
+++ b/lib/kredis/callbacks_proxy.rb
@@ -22,7 +22,7 @@ class Kredis::CallbacksProxy
 
     if CALLBACK_OPERATIONS[@type.class]&.include? method
       if @callback.respond_to? :call
-        @callback.call(@record)
+        @callback.call(@record || @type)
       elsif @callback.is_a? Symbol
         @record.send(@callback)
       end
--- a/lib/kredis/callbacks_proxy.rb
+++ b/lib/kredis/callbacks_proxy.rb
@@ -14,19 +14,15 @@ class Kredis::CallbacksProxy
     Kredis::Types::Slots => %i[ reserve release reset ]
   }
 
-  def initialize(type, record, callback)
-    @type, @record, @callback = type, record, callback
+  def initialize(type, callback)
+    @type, @callback = type, callback
   end
 
   def method_missing(method, *args, **kwargs, &block)
     result = @type.send(method, *args, **kwargs, &block)
 
     if CALLBACK_OPERATIONS[@type.class]&.include? method
-      if @callback.respond_to? :call
-        @callback.call(@record || @type)
-      elsif @callback.is_a? Symbol
-        @record.send(@callback)
-      end
+      @callback.call(@type)
     end
 
     result
--- a/lib/kredis/types.rb
+++ b/lib/kredis/types.rb
@@ -80,7 +80,7 @@ module Kredis::Types
   private
     def type_from(type_klass, config, key, after_change: nil, **options)
       type_klass.new(configured_for(config), namespaced_key(key), **options).then do |type|
-        after_change ? Kredis::CallbacksProxy.new(type, nil, after_change) : type
+        after_change ? Kredis::CallbacksProxy.new(type, after_change) : type
       end
     end
 end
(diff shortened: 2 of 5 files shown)


lib/kredis/types.rb
```diff
@@ -72,6 +72,12 @@ def slot(key, config: :shared)
   def slots(key, available:, config: :shared)
     Slots.new configured_for(config), namespaced_key(key), available: available
   end
+
+
+  def with_callback(type, key, **options)
+    callback = options.delete(:after_change)
+    Kredis::CallbacksProxy.new(Kredis.send(type, key, **options), nil, callback)
+  end
```
kaspth, line 80: Let's make an internal factory method to help out enriching the type if needed:

```ruby
  private
    def type_from(type_klass, config, key, after_change: nil, **options)
      type_klass.new(configured_for(config), namespaced_key(key), options).then do |type|
        after_change ? Kredis::CallbacksProxy.new(type, nil, after_change) : type
    end
  end
```

lib/kredis/callbacks_proxy.rb
```diff
@@ -22,7 +22,7 @@ def method_missing(method, *args, **kwargs, &block)
 
     if CALLBACK_OPERATIONS[@type.class]&.include? method
       if @callback.respond_to? :call
-        @callback.call(@record)
+        @callback.call(@record || @type)
```
kaspth, line 25: I know this looks like a straight shot but it makes it hard to reason about which paths this if/elsif are effected by.

Can we extract the `@record` requirement to just the attributes callsite and have it enrich the provided original callback with the record access?

kaspth (changes requested):
We've established that there's no extra method to use when adding a callback on the attributes side, e.g. `kredis_list :x, after_change: …` and not `kredis_with_callback :list, :x, after_change: …`, so let's follow that pattern here.

Such that users can do `Kredis.list :x, after_change: …` etc. See the internal factory suggestion.

lib/kredis/types.rb
julianrubisch: On it!

lib/kredis/callbacks_proxy.rb
julianrubisch: I'm afraid I don't quite follow. Could you use more/different words please?

kaspth: Sure, what I'm trying to get at is that the `CallbacksProxy`'s original contract is now being stretched into this new context — one where there isn't a record. I think we should follow that through by yanking the record state out of `CallbacksProxy` and use function composition to capture it instead and do something like:

```ruby
def kredis_connection_with(method, name, key, **options)
  ivar_symbol = :"@#{name}_#{method}"
  type = method.to_s.sub("kredis_", "")
  after_change = options.delete(:after_change)

  define_method(name) do
    else
      new_type = Kredis.send(type, kredis_key_evaluated(key) || kredis_key_for_attribute(name), **options)

      instance_variable_set ivar_symbol,
        after_change ? enrich_after_change_with_record_access(new_type, after_change) : new_type
    end
  end
end

# private method defined at the instance level:
def enrich_after_change_with_record_access(type, original_after_change)
  # May need a record = self here in case the scope capture can't rely on self.
  case original_after_change
  when Proc   then Kredis::CallbacksProxy.new(type, ->(_) { original_after_change.call(self) })
  when Symbol then Kredis::CallbacksProxy.new(type, ->(_) { send(original_after_change) })
  end
end
```

Note: this change also opts out of the proxy if there's no `after_change` such that we don't pay the `method_missing` cost needlessly.

julianrubisch: Ah, that makes a lot of sense now. Modified!

kaspth:
Sweet!

PR #37, opened 2021-07-21
Implement basic hash functionality

### LINKED ISSUE #28: Redis Hash support
Is there any plan to support `Redis Hash` datatype? if not already in-progress I can PR it.
`Hash` datatype is more memory-efficient to store small JSON data than an ordinary plain key-value.
Check [this](https://redis.io/topics/memory-optimization) section: `Using hashes to abstract a very memory efficient plain key-value store on top of Redis`

julianrubisch:
This PR introduces basic Redis native hash functionality to kredis. It proxies the following methods:

- `entries` => `hgetall`
- `set` => `hset`
- `get` => `hmget`, returning an array of values or a single value
- `keys` => `hkeys`
- `values`=> `hvals`

Note that all keys and values are treated as strings for now.

Closes #28

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,12 @@ hash.set("key" => "value", "key2" => "value2")        # => HSET myhash "key", "v
 %w[ key key2 ] == hash.keys                           # => HKEYS myhash
 %w[ value value2 ] == hash.values                     # => HVALS myhash
 
+high_scores = Kredis.hash "high_scores"
+high_scores.set(space_invaders: 100, pong: 42)                # HSET high_scores "space_invaders", "100", "pong", "42"
+%w[ space_invaders pong ] == high_scores.keys                 # HKEYS high_scores
+[100, 42] == high_scores.values                               # HVALS high_scores
+{ "space_invaders" => 100, "pong" => 42 } == high_scores.to_h # HGETALL high_scores
+
 head_count = Kredis.counter "headcount"
 0 == head_count.value              # => GET "headcount"
 head_count.increment               # => SET headcount 0 NX + INCRBY headcount 1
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -51,6 +51,13 @@ set << DateTime.tomorrow                                 # => SADD myset "2021-0
 2 == set.size                                            # => SCARD myset
 [ DateTime.tomorrow, DateTime.yesterday ] == set.members # => SMEMBERS myset
 
+hash = Kredis.set "myhash"
```
dhh, line 54: Shouldn't this be `Kredis.hash "myhash"`?

dhh:
This looks great! Should have an example and some tests for a typed hash as well.

julianrubisch:
Just so we're on the same page, what would your preferred implementation of a "typed" hash look like? 

- Typed values with string keys?
- Both typed as the same (doesn't make much sense imho)?

dhh:
I was thinking the keys would always be strings, and what you’d type is the values.

julianrubisch:
Awesome! I'll take a stab at this soon

julianrubisch:
Done! Would it make sense to add `with_indifferent_access` to `Kredis::Hash#entries` ?

dhh:
I like with_indifferent_access, but then we need to make sure it's done at all levels. For all get/set. And test that.

julianrubisch:
Affirmative, will add more tests. FWIW `set` should already work that way, as per the tests.

julianrubisch:
Added more tests for `with_indifferent_access`. Have I forgotten something?

kaspth:
Hey, thanks for this @julianrubisch — great base to start from and have a Hash that fits Ruby's semantics while allowing to use Redis as the store!

PR #41, opened 2021-08-19
Add remove method to hash type

LoranKloeze:
There was no way to delete a hash. This PR creates the `remove` method so that this works:
```ruby
hash = Kredis.hash "myhash"
hash.remove
```
This method calls `DEL myhash` in Redis

### DIFF
--- a/lib/kredis/types/hash.rb
+++ b/lib/kredis/types/hash.rb
@@ -1,7 +1,7 @@
 require "active_support/core_ext/hash"
 
 class Kredis::Types::Hash < Kredis::Types::Proxying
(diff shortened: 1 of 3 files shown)


lib/kredis/types/hash.rb
```diff
@@ -26,6 +26,9 @@ def delete(*keys)
     hdel types_to_strings(keys) if keys.flatten.any?
   end
 
+  def remove
```
marcoroth, line 29: Would it make sense to also alias `delete` to `remove` here?

kaspth:
Sweet, thanks!

PR #46, opened 2021-09-27
Add option to configure how the redis client is created

jorgemanrubia:
This adds a new option `config.kredis.connector` to customize how the `redis` client is created.

An example where this would be useful is configuring a proxy that wraps the Redis connection to provide failover capabilities (e.g: using makara).

cc @dhh

### DIFF
--- a/lib/kredis/connections.rb
+++ b/lib/kredis/connections.rb
@@ -3,7 +3,7 @@ require "redis"
 module Kredis::Connections
   mattr_accessor :connections, default: Hash.new
   mattr_accessor :configurator
-  mattr_accessor :connector, default: ->(config){ Redis.new(config) }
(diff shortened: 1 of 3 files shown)


lib/kredis/connections.rb
```diff
@@ -3,11 +3,12 @@
 module Kredis::Connections
   mattr_accessor :connections, default: Hash.new
   mattr_accessor :configurator
+  mattr_accessor :redis_connector, default: ->(config){ Redis.new(config) }
```
dhh, line 6: Would just go with `connector` – it's all redis here!

jorgemanrubia: 👍 `connector` it is, thanks!

PR #51, opened 2021-11-05
Add support for the NX (not exists) option to flags

valeksenko:
There are some Flag usage [Redis patterns](https://redis.io/commands/set#patterns) that benefit from having the flag set only if doesn't exist. SET support that use case with the `NX` option.

### DIFF
--- a/lib/kredis/types/flag.rb
+++ b/lib/kredis/types/flag.rb
@@ -3,8 +3,8 @@ class Kredis::Types::Flag < Kredis::Types::Proxying
 
   attr_accessor :expires_in
 
-  def mark(expires_in: nil)
-    set 1, ex: expires_in || self.expires_in
+  def mark(expires_in: nil, force: true)
+    set 1, ex: expires_in || self.expires_in, nx: !force
   end
 
   def marked?
--- a/lib/kredis/version.rb
+++ b/lib/kredis/version.rb
@@ -1,3 +1,3 @@
 module Kredis
-  VERSION = "0.4.0"
+  VERSION = "0.4.1"
 end
--- a/lib/kredis/version.rb
+++ b/lib/kredis/version.rb
@@ -1,3 +1,3 @@
 module Kredis
-  VERSION = "0.4.1"
+  VERSION = "0.4.0"
 end
--- a/README.md
+++ b/README.md
@@ -144,8 +144,8 @@ true == flag.marked?            # => EXISTS myflag
 flag.remove                     # => DEL myflag  
 false == flag.marked?           # => EXISTS myflag
 
-true == flag.mark(expires_in: 1.second, unless_exists: true)    #=> SET myflag 1 EX 1 NX
-false == flag.mark(expires_in: 10.seconds, unless_exists: true) #=> SET myflag 10 EX 1 NX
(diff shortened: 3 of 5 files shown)


lib/kredis/types/flag.rb
```diff
@@ -1,8 +1,8 @@
 class Kredis::Types::Flag < Kredis::Types::Proxying
   proxying :set, :exists?, :del
 
-  def mark(expires_in: nil)
-    set 1, ex: expires_in
+  def mark(expires_in: nil, unless_exists: false)
+    !!(set 1, ex: expires_in, nx: unless_exists)
```
valeksenko, line 5: have to use `!!` because with `nx: false` it returns `OK` vs boolean with `nx: true`

lib/kredis/version.rb
```diff
@@ -1,3 +1,3 @@
 module Kredis
-  VERSION = "0.4.0"
+  VERSION = "0.4.1"
```
kaspth, line 2: We'll change the version when we prepare a release, we shouldn't do this in this PR.

lib/kredis/types/flag.rb
kaspth: We don't guard against `"OK"` in any other client calls, we rely on it being truthy in Ruby, so let's just do:

```suggestion
    set 1, ex: expires_in, nx: unless_exists
```

kaspth (commented):
I'd like to see a use case for this from your app because just opening up to `nx` might not be what we want here.

For instance, we use [`nx: true` as a trick in `Counter`](https://github.com/rails/kredis/blob/8dd897480c3f6c9c7454a80589641fdb7634a295/lib/kredis/types/counter.rb#L8) to keep the original expiry and not reset on a subsequent call. We may want similar here.

valeksenko:
> I'd like to see a use case for this from your app because just opening up to `nx` might not be what we want here.
> 
> For instance, we use [`nx: true` as a trick in `Counter`](https://github.com/rails/kredis/blob/8dd897480c3f6c9c7454a80589641fdb7634a295/lib/kredis/types/counter.rb#L8) to keep the original expiry and not reset on a subsequent call. We may want similar here.

The use case is a flag to mark that we've communicated with a client in last week. We currently use the counter, since it's the only one supporting `nx` at the moment, but we really need a simple flag so the counter usage feels off since we discard the second return param of `.increment` just using the first param as an indication of a successful SET. I've used the same pattern plenty of time with a regular Redis usage so feels that Kredis would benefit from it.

dhh:
Let's flip the binary, so the attribute name can be in the affirmative:

```ruby
def mark(expires_in: nil, force: true)
  set 1, ex: expires_in, nx: !force
end
```

And then it would be nice to add this to the attribute definition too:

```
class User
  kredis_flag :banned, expires_in: 2.days, force: false
end
```

This would follow the addition of `expires_in` to the attribute specification in https://github.com/rails/kredis/pull/52.

valeksenko:
> Let's flip the binary, so the attribute name can be in the affirmative:

I like it better for sure. Renamed in https://github.com/rails/kredis/pull/51/commits/5d6bcd0427896cf7d56031d4b13b1e766af4c846

PR #54, opened 2021-11-23
Add []= and remove to hash callback methods

### LINKED ISSUE #53: Mutating a kredis_hash using []= doesn't trigger callbacks
```rb
class Foo < ApplicationRecord
  kredis_hash :bar, after_change: :touch_record

  def touch_record
    touch
  end
end
```

```rb
Foo.new.bar["test"] = "baz"
``` 

doesn't run the callback while

```rb
Foo.new.bar.update "test" => "baz"
```

does.

julianrubisch:
Here you go!

Btw took me only 4 minutes to add this. Sweet API ~~we~~ you've come up with here!

Closes #53


kaspth:
Sweet, thanks!

PR #56, opened 2021-12-02
Appending to unique list with limit should trim from the end

excid3:
When appending to a unique list with a limit, `ltrim` was clearing out the list every time.

> Out of range indexes will not produce an error: if start is larger than the end of the list, or start > end, the result will be an empty list (which causes key to be removed).

This changes it to calculate the limit from the end to preserve the list properly.

@kaspth

### DIFF
--- a/lib/kredis/types/unique_list.rb
+++ b/lib/kredis/types/unique_list.rb
@@ -16,7 +16,7 @@ class Kredis::Types::UniqueList < Kredis::Types::List
     multi do
       remove elements
       super
-      ltrim (limit - 1), -1 if limit
+      ltrim -limit, -1 if limit
     end if Array(elements).flatten.any?
   end
   alias << append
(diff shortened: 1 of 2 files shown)


lib/kredis/types/unique_list.rb
```diff
@@ -16,7 +16,7 @@ def append(elements)
     multi do
       remove elements
       super
-      ltrim (limit - 1), -1 if limit
+      ltrim -limit, -1 if limit
```
kaspth, line 19: Interesting that going from the tail end isn't zero-indexed, e.g. the base is `-1` and not `-0`.

kaspth (approved):

lib/kredis/types/unique_list.rb
excid3: I think it matches Ruby with ranges:

```ruby
irb(main):001:0> [1,2,3,4][-3..-1]
=> [2, 3, 4]
```

PR #58, opened 2021-12-03
Deduplicate new elements for unique lists

excid3:
Unique lists validates the new elements were not an empty list, but does not check for duplicate items in the array being passed in.

This makes removes any duplicates before passing them to Redis.


kaspth:
Discussed with @excid3 to remove the rare and untested flatten handling and let callers handle that.

PR #68, opened 2022-02-18
Use block parameter to pipeline in Redis#multi

capripot:
In preparation of Redis 5, as info message alerts.

> Pipelining commands on a Redis instance is deprecated and will be removed in Redis 5.0.0.
>
> redis.multi do
>   redis.get("key")
> end
>
> should be replaced by
>
> redis.multi do |pipeline|
>   pipeline.get("key")
> end
>
> (called from /app/vendor/bundle/ruby/2.7.0/gems/kredis-1.0.1/lib/kredis/types/proxy.rb:13:in `multi'}>

Using block parameter is compatible down to [`redis` gem version 3.0](https://www.rubydoc.info/gems/redis/3.0.0/Redis#multi-instance_method), current dependency to `redis` is defined as `~> 4.2`.

### DIFF
--- a/lib/kredis/types/counter.rb
+++ b/lib/kredis/types/counter.rb
@@ -1,19 +1,19 @@
 class Kredis::Types::Counter < Kredis::Types::Proxying
-  proxying :multi, :set, :get, :del, :exists?
+  proxying :multi, :set, :incrby, :decrby, :get, :del, :exists?
 
   attr_accessor :expires_in
 
   def increment(by: 1)
     multi do |pipeline|
-      pipeline.set key, 0, ex: expires_in, nx: true
-      pipeline.incrby key, by
+      pipeline.set 0, ex: expires_in, nx: true
+      pipeline.incrby by
     end
   end
 
   def decrement(by: 1)
     multi do |pipeline|
-      pipeline.set key, 0, ex: expires_in, nx: true
-      pipeline.decrby key, by
+      pipeline.set 0, ex: expires_in, nx: true
+      pipeline.decrby by
     end
   end
 
--- a/lib/kredis/types/counter.rb
+++ b/lib/kredis/types/counter.rb
@@ -1,19 +1,19 @@
 class Kredis::Types::Counter < Kredis::Types::Proxying
-  proxying :multi, :set, :incrby, :decrby, :get, :del, :exists?
+  proxying :multi, :set, :get, :del, :exists?
 
   attr_accessor :expires_in
 
   def increment(by: 1)
     multi do |pipeline|
-      pipeline.set 0, ex: expires_in, nx: true
-      pipeline.incrby by
+      pipeline.set key, 0, ex: expires_in, nx: true
+      pipeline.incrby key, by
     end
   end
 
   def decrement(by: 1)
     multi do |pipeline|
-      pipeline.set 0, ex: expires_in, nx: true
-      pipeline.decrby by
+      pipeline.set key, 0, ex: expires_in, nx: true
+      pipeline.decrby key, by
     end
   end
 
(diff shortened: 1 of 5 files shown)


dhh:
Looks like tests are failing?

capripot:
Ha true! My bad. I just fixed the calls adding the key to them, now `bin/test` is all green! ✅

lib/kredis/types/unique_list.rb
```diff
@@ -1,28 +1,28 @@
...
   attr_accessor :typed, :limit
 
   def prepend(elements)
     elements = Array(elements).uniq
     return if elements.empty?
 
-    multi do
-      remove elements
+    multi do |pipeline|
+      types_to_strings(elements, typed).each { |element| pipeline.lrem key, 0, element }
```
capripot, line 12: It repeats `List#remove`'s code, but making `remove` accept a pipeline looked equally unideal.

lib/kredis/types/counter.rb
```diff
@@ -1,19 +1,19 @@
 class Kredis::Types::Counter < Kredis::Types::Proxying
-  proxying :multi, :set, :incrby, :decrby, :get, :del, :exists?
+  proxying :multi, :set, :get, :del, :exists?
```
dhh, line 2: Removing these two from the proxy would be a breaking API change. Potentially someone were calling these directly. Think we need to leave them exposed.

lib/kredis/types/set.rb
```diff
@@ -18,9 +18,9 @@ def remove(*members)
   end
 
   def replace(*members)
-    multi do
-      del
-      add members
+    multi do |pipeline|
+      pipeline.del key
+      pipeline.sadd key, types_to_strings(members, typed) if members.flatten.any?
```
dhh, line 23: Need to extract `types_to_strings(members, typed) if members.flatten.any?` into a method. It's repeated three times.

lib/kredis/types/unique_list.rb
dhh: Man, this is a pretty nasty API regression. Need to take a look and see what else we can come up with.

capripot:
Here is what I came up with:
- using a new proxy within `Kredis#multi` as `Kredis::Types::Proxy.new(pipeline, key)`, it has the advantage of having same interface as with `redis`
- for "shared" methods, I used an optional keyword argument

We could also have an optional keyword argument on all proxied methods, that would create less objects, but could be a little inelegant as we would need to pass `pipeline: pipeline` everywhere, beyond "shared" methods.
```ruby
class Kredis::Types::Proxy
  def method_missing(method, *args, **kwargs)
    Kredis.instrument :proxy, **log_message(method, *args, **kwargs) do
      failsafe do
        (kwargs.delete(:pipeline) || redis).public_send method, key, *args, **kwargs
      end
    end
  end
```

lib/kredis/types/set.rb
```diff
@@ -8,19 +8,19 @@ def members
...
-  def add(*members)
-    sadd types_to_strings(members, typed) if members.flatten.any?
+  def add(*members, pipeline: nil)
+    (pipeline || proxy).sadd types_to_strings(members, typed) if members.flatten.any?
   end
   alias << add
 
-  def remove(*members)
-    srem types_to_strings(members, typed) if members.flatten.any?
+  def remove(*members, pipeline: nil)
```
capripot, line 16: For consistency

lewispb:
Played with this and came up with something a little simpler: https://github.com/rails/kredis/pull/78

dhh:
Going to merge this for now. @lewispb we can take a swing at the ivar version if we determine that's a thread safe approach afterwards.

Issue #62, opened 2022-01-21
Counter increment and decrement return value


pbstriker38:
Current return value

```ruby
irb(main):001:0> counter = Kredis.counter('my_counter', expires_in: 1.hour)
irb(main):002:0> counter.increment
=> [true, 1]
irb(main):003:0> counter.decrement
=> [false, 0]
```

expected return value

```ruby
irb(main):001:0> counter = Kredis.counter('my_counter', expires_in: 1.hour)
irb(main):002:0> counter.increment
=> 1
irb(main):003:0> counter.decrement
=> 0
```

I can open a pull request if this is an acceptable change.

dhh:
Yes please!

PR #72, opened 2022-02-24
Return counter value after incrementing/decrementing

pbstriker38:
- Redis multi return an array with the responses of each command that
 was run. Use [-1] to get the last value of the array and return it.

closes: #62

### DIFF
--- a/lib/kredis/types/counter.rb
+++ b/lib/kredis/types/counter.rb
@@ -7,14 +7,14 @@ class Kredis::Types::Counter < Kredis::Types::Proxying
     multi do
       set 0, ex: expires_in, nx: true
       incrby by
-    end
+    end[-1]
   end
 
   def decrement(by: 1)
     multi do
       set 0, ex: expires_in, nx: true
       decrby by
-    end
+    end[-1]
   end
 
   def value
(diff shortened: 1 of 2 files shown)


pbstriker38:
I wasn't sure if we wanted to go for readability or performance. I can switch this to use `.last` if it's more clear.

```ruby
response = [false, 4]
Benchmark.ips do |x|
  x.report("[1]")  { response[1] }
  x.report("[-1]")  { response[-1] }
  x.report("last") { response.last }
  x.compare!
end
Warming up --------------------------------------
                 [1]     2.138M i/100ms          
                [-1]     2.058M i/100ms          
                last     1.857M i/100ms          
Calculating -------------------------------------
                 [1]     21.546M (± 2.6%) i/s -    109.040M in   5.064581s
                [-1]     21.314M (± 2.2%) i/s -    107.028M in   5.024219s
                last     18.051M (± 5.1%) i/s -     90.974M in   5.054999s
                                                 
Comparison:                                      
                 [1]: 21546133.6 i/s
                [-1]: 21313633.8 i/s - same-ish: difference falls within error
                last: 18050678.0 i/s - 1.19x  (± 0.00) slower
```

jeremy (approved):
`#last` does read better, but this works. Thanks @pbstriker38!

Issue #66, opened 2022-02-08
Default proc if empty


tleish:
Is there a way to set a default value for a list using a proc?  I'm thinking something like:

```ruby
class User < ApplicationRecord
  kredis_list :role_ids, default: ->(user) { user.roles.pluck(:id) }
end
```

Then if there's a nil value (either because it was never created or expired at some point), it recreates itself with the latest data.

I can accomplish the same thing with:

```ruby
class User < ApplicationRecord
  kredis_list :role_ids_cache, default: ->(user) { user.roles.pluck(:id) }

  def role_ids
    self.role_ids_cache.append(self.roles.pluck(:id)) unless self.role_ids_cache.elements.empty?
    self.role_ids_cache
  end
end
```

... but I prefer dealing with one method/variable vs two.

dhh:
I like the idea of this. Please do explore as a PR 👍

PR #73, opened 2022-02-24
Add support for list default value

mwsteb:
First of all, thanks for a great gem!

Here's my attempt at implementing #66.

I wasn't sure how strict to be about default value type. For now, I opted for giving the user scissors and hoping they won't run with them.

Hope you don't mind I butted in like that, @tleish! Not sure if you started to work on this or not.

closes: https://github.com/rails/kredis/issues/66

### DIFF
--- a/lib/kredis/attributes.rb
+++ b/lib/kredis/attributes.rb
@@ -42,8 +42,8 @@ module Kredis::Attributes
       kredis_connection_with __method__, name, key, config: config, after_change: after_change, expires_in: expires_in
     end
 
-    def kredis_list(name, key: nil, typed: :string, config: :shared, after_change: nil)
-      kredis_connection_with __method__, name, key, typed: typed, config: config, after_change: after_change
+    def kredis_list(name, key: nil, typed: :string, config: :shared, after_change: nil, default: nil)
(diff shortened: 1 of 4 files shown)


tleish:
@mwsteb - I don't mind at all.  While my example  in #66 shows `kredis_list`, I want this with other (if not all) Kredis types.  I might suggest adding more default types to this PR, or removing `closes: #66` from the PR comment so additional PRs can be added with default types.

mwsteb:
Makes sense. Since it's not a lot of code and it's all related, I can do it all in this PR.

I'm not sure about some types:
* Proxy - unclear what this type is exactly and how it's used and what a typical default value would be 
* Cycle - setting the default value can be achieved by manipulating the order of the list - eg. `Kredis.cycle "mycycle", values: %i[ three one two ]`. Is adding an additional `default` option worth it?
* Slot(s) - seems a bit contrived to be using default value for this type, but maybe I just lack imagination :)

tleish:
Proxy is a base class, so I don't think you need to worry about it.   I assume it's safe to skip defaults for the other types as well for now.

dhh:
Superseded by #89

Issue #74, opened 2022-02-25
Use a sorted set for unique list functionality


lewispb:
I benchmarked an implementation of a unique list using a [Redis Sorted Set](https://redis.io/topics/data-types#sorted-sets) as opposed to the current implementation which uses a Redis list.

https://gist.github.com/lewispb/bcb190cf72c93b67f57671e0287fa7a7

I think this shows potential performance improvements of at least 5x, or depending on the size of the list more than that.

If we want to move towards using a sorted set, we'd need to think about the migration. Maybe we maintain unique list for backwards compatiblity but add a new data structure, `Kredis.sorted_set`?

dhh:
Think the problem is that redis already has a sorted set with certain expectations tied to it (like being unordered). Think we might have to go with something like unique_fast_list, deprecate the old one, then set a cut-off for passing over.

Although I suppose it'd be even nicer if we could detect the old data format and convert?

lewispb:
> Although I suppose it'd be even nicer if we could detect the old data format and convert?

I think this would be possible actually. We could rescue the error: `WRONGTYPE Operation against a key holding the wrong kind of value (Redis::CommandError)`. Then there'd be a one-off hit to migrate the data there and then from a list to a sorted set.

dhh:
Seems worth exploring.

lewispb:
👍 I'll look into it

PR #76, opened 2022-02-25
Improved version of UniqueList: OrderedSet

lewispb:
Closes https://github.com/rails/kredis/issues/74

Using a [benchmark script](https://gist.github.com/lewispb/577db0cf4c75a23f6e07101df3c3488b), we can see the improvement in performance by using a Sorted Set in some scenarios:

```
              user       system     total    real
ordered_set 0.131565   0.018613   0.150178 (  0.174106)
unique_list 1.109980   0.133186   1.243166 (  4.544614)
```

Initially, we can introduce this as a new datatype. But to follow on, a migration mechanism will be introduced so that we can replace unique_list.


lib/kredis/types/unique_list_legacy.rb
```diff
@@ -0,0 +1,29 @@
+# You'd normally call this a set, but Redis already has another data type for that
+class Kredis::Types::UniqueListLegacy < Kredis::Types::List
```
lewispb, line 2: Renamed the existing `UniqueList` to `UniqueListLegacy`.

lib/kredis/types/unique_list.rb
```diff
@@ -1,29 +1,84 @@
...
+      end
+
+      multi do
+        zadd(elements_with_scores)
+        trim(from_beginning: prepending)
+      end
+    end
+
+    def current_nanoseconds(negative:)
+      "%10.9f" % (negative ? -Time.now.to_f : Time.now.to_f)
```
lewispb, line 49: Use monotonic clock - `Process.clock_gettime(Process::CLOCK_MONOTONIC)`

lib/kredis/types/unique_list.rb
```diff
@@ -1,29 +1,84 @@
...
+    def insert(elements, prepending: false)
+      elements = Array(elements)
+      return if elements.empty?
+
+      elements_with_scores = types_to_strings(elements, typed).map do |element|
+        [ current_nanoseconds(negative: prepending), element ]
+      end
+
+      multi do
+        zadd(elements_with_scores)
```
lewispb, line 43: Consider [LT / GT options](https://redis.io/commands/ZADD#zadd-options) for ordering safety, avoiding problems with clock skew.

lib/kredis/types/unique_list.rb
```diff
@@ -1,29 +1,93 @@
...
+      end
+    end
+
+    def generate_score(index, negative:)
+      score = Integer([ *basis_score_components, index.to_s.rjust(6, "0") ].join, 10)
+      negative ? -score : score
+    end
+
+    def basis_score_components
+      [ redis_time + Time.now.utc.to_i, Time.now.utc.nsec ]
```
lewispb, line 54: Todo: further testing and rework to account for DST etc.

lib/kredis/types/unique_list.rb
```diff
@@ -1,29 +1,96 @@
...
+    end
+
+    def generate_base_score(negative:)
+      current_time = process_start_time + process_uptime
+
+      negative ? -current_time : current_time
+    end
+
+    def process_start_time
+      @process_start_time ||= redis.time.join(".").to_f - process_uptime
```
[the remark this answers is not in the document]
lewispb: Do you envisage any problems if this were used in such a setup?

lib/kredis/types/unique_list.rb
```diff
@@ -1,29 +1,96 @@
...
+    end
+
+    def process_uptime
+      Process.clock_gettime(Process::CLOCK_MONOTONIC)
```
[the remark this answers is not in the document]
lewispb: I think Process.clock_gettime is OK in JRuby 9 +

https://blog.dnsimple.com/2018/03/elapsed-time-with-ruby-the-right-way/

lib/kredis/types/unique_list.rb
```diff
@@ -1,29 +1,96 @@
...
+    end
+
+    def trim(from_beginning:)
+      return unless limit&.positive?
```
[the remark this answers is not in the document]
lewispb: Good idea!

lewispb:
@jeremy thanks for all the feedback. This is much simpler now. I like the idea of splitting out the migration into a follow-up PR.

lib/kredis/types/ordered_set.rb
```diff
@@ -0,0 +1,71 @@
...
+    @limit = limit
+  end
+
+  private
+    def insert(elements, prepending: false)
+      elements = Array(elements)
+      return if elements.empty?
+
+      elements_with_scores = types_to_strings(elements, typed).map.with_index do |element, index|
+        score = generate_base_score(negative: prepending) + index / 100000
```
intrip, line 37: Could you elaborate what `index / 10000` is for?

lib/kredis/types/ordered_set.rb
```diff
@@ -0,0 +1,71 @@
...
+        pipeline.zadd(elements_with_scores)
+        trim(from_beginning: prepending, pipeline: pipeline)
+      end
+    end
+
+    def generate_base_score(negative:)
+      current_time = process_start_time + process_uptime
+
+      negative ? -current_time : current_time
+    end
```
intrip, line 52: Clever approach 👍!  

Just a note for your consideration: 

What would happen in case of skewed Redis clocks / concurrent writes? Is possible to face the scenario where appending 2 arrays would lead to mixing their elements?

```ruby
# In process A
ordered_set.append([1, 2, 3])
# In process B
ordered_set.append([4, 5, 6])
ordered_set.elements
# Is this possible?
# [1, 2, 4, 3, 5, 6]
```

Maybe a less clever approach that relies getting the max/min (with an optimistic lock) would be more resilient?

```
WATCH zset
# get max
element = ZRANGE zset -1 -1
# compute scores max +1 for each element
MULTI
# add elements
ZADD zset elemens
# trim
EXEC
```

intrip (commented):
@lewispb Great approach! I've just left a note for your consideration

lib/kredis/types/ordered_set.rb
lewispb: > Redis sorted sets use a double 64-bit floating point number to represent the score. In all the architectures we support, this is represented as an IEEE 754 floating point number

I was trying to store the maximum unique precision possible in Redis, to store scores (times). By dividing a large integer (seconds) by 100000 we get a float. e.g:

```rb
>> Time.now.to_i
=> 1653898149
>> Time.now.to_f / 100000
=> 16538.98132595902
>> 3.times { p Time.now.to_f / 100000 }
16538.98262287975
16538.982622880398
16538.98262288052
```

Even though we lose some precision, each iteration produces a higher score than the previous one, perfect for our use case of appending.

jeremy (approved):
Looking great.

Let's consider how to guide folks to the improved data structure. Perhaps remove the unique list example from the README, or move to the bottom and mark as deprecated? Could do this in a separate PR, since it also begs "well, how do I migrate?"

PR #78, opened 2022-03-01
Avoid having to pass a Redis pipeline argument around

lewispb:
Improves upon https://github.com/rails/kredis/pulls/68 by simplifying the internal implementation, avoiding passing a pipeline as a method argument

### DIFF
--- a/lib/kredis/types/proxy.rb
+++ b/lib/kredis/types/proxy.rb
@@ -4,6 +4,8 @@ class Kredis::Types::Proxy
 
   attr_accessor :redis, :key
 
+  thread_mattr_accessor :pipeline
+
   def initialize(redis, key, **options)
     @redis, @key = redis, key
     options.each { |key, value| send("#{key}=", value) }
@@ -11,17 +13,17 @@ class Kredis::Types::Proxy
 
   def multi(*args, **kwargs, &block)
     redis.multi(*args, **kwargs) do |pipeline|
-      Thread.current[:pipeline] = pipeline
+      self.pipeline = pipeline
       block.call
     ensure
-      Thread.current[:pipeline] = nil
+      self.pipeline = nil
     end
   end
 
   def method_missing(method, *args, **kwargs)
     Kredis.instrument :proxy, **log_message(method, *args, **kwargs) do
       failsafe do
-        (Thread.current[:pipeline] || redis).public_send method, key, *args, **kwargs
+        (pipeline || redis).public_send method, key, *args, **kwargs
       end
     end
   end
--- a/lib/kredis/types/proxy.rb
+++ b/lib/kredis/types/proxy.rb
@@ -2,7 +2,7 @@ class Kredis::Types::Proxy
   require_relative "proxy/failsafe"
   include Failsafe
 
-  attr_accessor :redis, :key, :pipeline
+  attr_accessor :redis, :key
 
   def initialize(redis, key, **options)
     @redis, @key = redis, key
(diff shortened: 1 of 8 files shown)


jeremy:
Related: #68

dhh:
@Jeremy Thoughts on the thread safety issue here? Seems to me that #68 does not have that problem?

lewispb:
Updated to use a thread-local variable. Is that sufficient?

lib/kredis/types/proxy.rb
```diff
@@ -9,19 +9,19 @@ def initialize(redis, key, **options)
...
-    return redis.multi unless block
-
-    redis.multi do |pipeline|
-      block.call(Kredis::Types::Proxy.new(pipeline, key))
+  def multi(*args, **kwargs, &block)
+    redis.multi(*args, **kwargs) do |pipeline|
+      Thread.current[:pipeline] = pipeline
+      block.call
+    ensure
+      Thread.current[:pipeline] = nil
```
jorgemanrubia, line 17: Consider using a `thread_mattr_accessor` so that it takes care of using a key based on the attribute name and the class name. If not, I'd try to make `:pipeline` more specific: e.g: :kredis_pipeline or something like that.

jorgemanrubia (commented):
Nice solution @lewispb! Minor suggestion below but I love how this simplifies things.

I think that with the thread-based attribute this code is thread-safe now.

lib/kredis/types/proxy.rb
```diff
@@ -1,27 +1,31 @@
 class Kredis::Types::Proxy
+  require "active_support/core_ext/module"
```
jorgemanrubia, line 2: I'd move the require to `lib/kredis.rb`:

https://github.com/rails/kredis/blob/2469a8d6f166ec6b4331f339eb128a3dcf315945/lib/kredis.rb#L2

jorgemanrubia (approved):
Thanks @lewispb, one minor comment below regarding the require.

skipkayhil:
Does this not throw deprecation warnings in the latest version of redis-rb? I thought the pipeline block argument was basically required now

lib/kredis/types/proxy.rb
```diff
@@ -4,24 +4,26 @@ class Kredis::Types::Proxy
...
+    ensure
+      self.pipeline = nil
     end
   end
 
   def method_missing(method, *args, **kwargs)
     Kredis.instrument :proxy, **log_message(method, *args, **kwargs) do
       failsafe do
-        redis.public_send method, key, *args, **kwargs
+        (pipeline || redis).public_send method, key, *args, **kwargs
```
jeremy, line 26: Noting that Proxying subclasses that (mistakenly, perhaps) talk to `redis` directly will need to take care to check for `pipeline` presence as well.

Perhaps we can coalesce "current pipeline or redis" into the `redis` method itself, e.g.

```diff
diff --git a/lib/kredis/types/proxy.rb b/lib/kredis/types/proxy.rb
index 172ff2f..12dc26d 100644
--- a/lib/kredis/types/proxy.rb
+++ b/lib/kredis/types/proxy.rb
@@ -2,7 +2,9 @@ class Kredis::Types::Proxy
   require_relative "proxy/failsafe"
   include Failsafe
 
-  attr_accessor :redis, :key
+  attr_accessor :key
+
+  thread_mattr_accessor :pipeline
 
   def initialize(redis, key, **options)
     @redis, @key = redis, key
@@ -27,6 +29,10 @@ class Kredis::Types::Proxy
   end
 
   private
+    def redis
+      pipeline || @redis
+    end
+
     def log_message(method, *args, **kwargs)
       args      = args.flatten.reject(&:blank?).presence
       kwargs    = kwargs.reject { |_k, v| v.blank? }.presence
```

I'd _think_ we could remove the `:redis` attr from Proxying, too. It's a bit of a foot-gun: it can't be safely called without knowledge of whether there's an open MULTI transaction—knowledge which is not available to Proxying subclasses or their callers:
```diff
diff --git a/lib/kredis/types/proxying.rb b/lib/kredis/types/proxying.rb
index a7c0e55..3607f0f 100644
--- a/lib/kredis/types/proxying.rb
+++ b/lib/kredis/types/proxying.rb
@@ -1,14 +1,14 @@
 require "active_support/core_ext/module/delegation"
 
 class Kredis::Types::Proxying
-  attr_accessor :proxy, :redis, :key
+  attr_accessor :proxy, :key
 
   def self.proxying(*commands)
     delegate *commands, to: :proxy
   end
 
   def initialize(redis, key, **options)
-    @redis, @key = redis, key
+    @key = key
     @proxy = Kredis::Types::Proxy.new(redis, key)
     options.each { |key, value| send("#{key}=", value) }
   end
```

lewispb: Those further changes make good sense to me, thanks @jeremy.

Issue #77, opened 2022-02-28
Target Redis server version


lewispb:
While working on a feature recently I was exploring the use of the `ZADD` options `GT` and `LT`. These are [only available in Redis 6.2 onwards](https://redis.io/commands/ZADD#history).

Do we have a minimum Redis server version in mind while developing Kredis? If not, can we document one?

The only reference to a version I could see was this line: https://github.com/rails/kredis/blob/main/lib/kredis/migration.rb#L8

dhh:
I believe we're at Redis 4+ at the moment since that's what Rails suggest in the default Gemfile.

But we probably could just gate new features on a redis version check? Then fail if not available?

lewispb:
Added a note to the readme in https://github.com/rails/kredis/pull/79.

I backed away from `GT` / `LT` so nothing in mind that would require gating right now, but good to know that's an option for the future.

Might be worth also adding a matrix of supported Redis server versions, say 4 and 6.2, and client libraries to CI?

PR #79, opened 2022-03-01
Note tested / supported versions of Redis in the Readme

lewispb:
Relates to https://github.com/rails/kredis/issues/77

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby: [2.7, "3.0"]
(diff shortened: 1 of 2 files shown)


dhh:
On second reflection, it's a bit of a weak guarantee, this? "tested on". Maybe we should just find out where it doesn't work, and document the minimum viable version?

lewispb:
Yeah, thinking of putting together a CI matrix of Redis clients + servers, if that sounds good?

lewispb:
Redis server version 2 and 3 cause failures in the test suite, so we can say that for 100% of Kredis functionality to work, Redis server version 4 onwards is required.

I've updated the Readme to reflect this, and added a CI matrix for Redis server versions (if you could approve the workflow, that'd be great).

PR #86, opened 2022-05-31
Support expires_in timeout for Lists / UniqueLists

swanson:
Similar to https://github.com/rails/kredis/pull/19

Adds support for expiring lists (and unique lists).

You specify a "timeout" via `expires_in` when defining the attribute. Adding or removing items from the list bumps the timeout. This is due to how Redis works internally -- calling `lpush` or `rpush` on a key will clear any existing expiry, so after the list has been modified, we have to resend the `expire` command.

## Motivation

We have a use-case for Kredis where we want to store a list of IDs for items that a user can collapse/expand. But after a few hours/days, we don't need to keep persisting the state. This prevents our Redis storage space from expanding indefinitely as old lists will eventually expire.

```rb
class Users::PlanState
  include Kredis::Attributes
  
  kredis_list :expanded_item_ids, expires_in: 3.days

  attr_accessor :user, :plan

  def id
    "users:#{user.id}:plans:#{plan.id}"
  end

  def item_expanded?(item)
    expanded_item_ids.include?(item.id)
  end
end
```

Without this change, we have to workaround the inability to expire lists by using the `kredis_json` type to store an array.

A similar example: using Kredis to store user-specified filters. You may want to persist these filters across page visits/reloads but after some time has elapsed, it would be acceptable to "forget" the selection.

image

### DIFF
--- a/lib/kredis/types/list.rb
+++ b/lib/kredis/types/list.rb
@@ -10,21 +10,26 @@ class Kredis::Types::List < Kredis::Types::Proxying
 
   def remove(*elements, pipeline: nil)
     types_to_strings(elements, typed).each { |element| (pipeline || proxy).lrem 0, element }
-    (pipeline || proxy).expire expires_in.to_i if expires_in
+    refresh_expiration(pipeline || proxy)
   end
 
   def prepend(*elements, pipeline: nil)
     (pipeline || proxy).lpush types_to_strings(elements, typed) if elements.flatten.any?
-    (pipeline || proxy).expire expires_in.to_i if expires_in
+    refresh_expiration(pipeline || proxy)
   end
 
   def append(*elements, pipeline: nil)
     (pipeline || proxy).rpush types_to_strings(elements, typed) if elements.flatten.any?
-    (pipeline || proxy).expire expires_in.to_i if expires_in
+    refresh_expiration(pipeline || proxy)
   end
   alias << append
 
   def clear
     del
   end
+
+  private
+    def refresh_expiration(target)
+      target.expire expires_in.to_i if expires_in
+    end
 end
--- a/lib/kredis/types/list.rb
+++ b/lib/kredis/types/list.rb
@@ -10,17 +10,17 @@ class Kredis::Types::List < Kredis::Types::Proxying
 
   def remove(*elements, pipeline: nil)
     types_to_strings(elements, typed).each { |element| (pipeline || proxy).lrem 0, element }
-    expire expires_in.to_i if expires_in
+    (pipeline || proxy).expire expires_in.to_i if expires_in
   end
 
   def prepend(*elements, pipeline: nil)
     (pipeline || proxy).lpush types_to_strings(elements, typed) if elements.flatten.any?
-    expire expires_in.to_i if expires_in
+    (pipeline || proxy).expire expires_in.to_i if expires_in
   end
 
   def append(*elements, pipeline: nil)
     (pipeline || proxy).rpush types_to_strings(elements, typed) if elements.flatten.any?
-    expire expires_in.to_i if expires_in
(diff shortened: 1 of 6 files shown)


swanson:
It looks like I may need special handling for the case of `pipeline || proxy` -- will investigate further.

swanson:
> It looks like I may need special handling for the case of `pipeline || proxy` -- will investigate further.

Updated the PR to support `unique_list` and address the need for `pipeline || proxy`

swanson:
This is ready for a review (@dhh would you mind taking a look?)

pedrofs:
Hi, @swanson,

I also need to expire a hash. As the `expire` and `expireat` are capabilities for any redis key, it makes sense to go a step further and move the expire logic to a module and make it available for every type, what do you think?

lib/kredis/types/list.rb
```diff
@@ -10,14 +10,17 @@ def elements
 
   def remove(*elements, pipeline: nil)
     types_to_strings(elements, typed).each { |element| (pipeline || proxy).lrem 0, element }
+    (pipeline || proxy).expire expires_in.to_i if expires_in
```
jorgemanrubia, line 13: You can consider extracting a `refresh_expiration` method to reuse in the three invocations:

```ruby
def prepend(*elements, pipeline: nil)
  (pipeline || proxy).lpush types_to_strings(elements, typed) if elements.flatten.any?
  refresh_expiration(pipeline)
end

# ...

def refresh_expiration(pipeline)
  (pipeline || proxy).expire expires_in.to_i if expires_in
end
```

test/types/list_test.rb
```diff
@@ -41,6 +41,26 @@ class ListTest < ActiveSupport::TestCase
...
+
+    assert_equal %w[ 1 2 3 ], @list.elements
+
+    sleep 1.1.seconds
+
+    assert_equal [], @list.elements
+
+    @list.prepend(4)
+
+    sleep 0.6.seconds
```
jorgemanrubia, line 56: I understand the usage of active delays to test this, but maybe you can replace them with using Redis' `#ttl` method, to base the assertions on having the expected TTL is set for the key. I think it's worth avoiding active delays in tests unless there is no other option (they can be brittle and slow down the suite).

jorgemanrubia (commented):
This is a good addition @swanson, thanks. 

As @pedrofs says, it would be good to extend this to other attributes, but we can generalize in a separated PR. Also, that `pipeline` addition from #68 made code a little bit messier, with that `pipeline` param hanging around. But again that's another concern we can tackle separately. This PR is a valuable improvement 👍.

test/types/list_test.rb
swanson: I agree; I followed the pattern for the other expiration-related tests and since the whole suite runs in 15 seconds it didn't seem pressing. Would you like me to tackle that as part of this PR?

jorgemanrubia: I see. Yes, I think it would be good to start changing the pattern in these tests, instead of extending the existing approach further.

swanson: I updated the unit tests to use `ttl` instead of sleep. As long as the assert comes within one-second for setting it, it seems to work ok (`ttl` has a precision of seconds) and avoids the sleeps.

I left the sleeps in the `attributes_test` as that seems more appropriate to not test the internals of the kredis types.

jorgemanrubia:
@swanson thanks a lot. Yes, with #91 in place I lean towards pursuing that approach and get some generic support here instead. Thanks a lot for your work on this PR in any case, you showed the need and triggered the work for the final solution 👍.

Please close this one. I'm noticing I don't have permissions in this repo to close PRs yet (I need to get that sorted out) 😅.

PR #87, opened 2022-06-05
Implement a key search interface through Kredis.search

LoranKloeze:
At this moment there is no functionality in Kredis to search for keys. I would like to propose to implement a key search functionality. The functionality abstracts the Redis SCAN command away from the user. 

## Kredis.search
The search function is implemented in `Kredis.search`. It respects `Kredis.namespace` if set.

### Normal usage
Suppose we cache user names via Kredis using `User.first(500).each {|user| Kredis.string("user_names:#{user.id}").value = user.name }`. We want to fetch all user names using Kredis.
```ruby
user_names = []
Kredis.search('user_names:*') do |key_batch| # key_batch.size is approx. 1000 keys
  user_names += key_batch.map { |key| Kredis.string(key).value }
end

(pry)> user_names
=> ["john", "marie", "alice", "bob", ...]
```

### Different batch size
```ruby
...
Kredis.search('user_names:*', batch_size: 200) do |key_batch| # key_batch.size is approx. 200 keys
...
```

### Different config
```ruby
...
Kredis.search('user_names:*', config: :secondary) { ... } # use config/redis/secondary.yml
...
```

## Business case

- It's an easy to use interface
- Users don't have to fall back to raw Redis to scan for keys

### Todo
- The interface still feels a little unruby, especially the disability to map the keys directly into an array feels clunky. Maybe the abstraction should be taken to a higher level? 

Any feedback is very much appreciated! My main doubt: is this something Kredis should take care of? Does it adhere to the spirit of Kredis being a 'keyed' Redis or should users just use raw Redis?

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,24 @@ class Person < ApplicationRecord
 end
 ```
 
+There is an easy way to search for keys:
+
+```ruby
+# Returns the search results in batches of *approximately* 1.000 keys
+Kredis.search('user_names:*') do |key_batch|
+  key_batch == [ 'user_names:1', 'user_names:2', ... ]
+end
+
+# Optionally you can change the batch size hint. Since it's a hint,
+# there is no guarantee the size of each batch matches the hint.
+#
+# Returns the search results in batches of *approximately* 750 keys 
+Kredis.search('user_names:*', batch_size: 750) { ... }
+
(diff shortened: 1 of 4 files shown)


LoranKloeze:
After some feedback I'm going to close this issue. From the feedback it seems this functionality has a better fit in a separate gem. 🙂

PR #89, opened 2022-06-11
Default values for entries

### LINKED ISSUE #66: Default proc if empty

tleish:
fixes #66

Default values can be static or Proc.  

Types with default:
- list
- uniq_list
- string
- integer
- decimal
- datetime
- float
- set
- proxy
- json
- counter
- string
- hash
- boolean

Types without default:
- cycle
- enum
- flag
- slots

TODO:
Looking for feedback.  After getting feedback I can update the README to include examples of using default


lib/kredis/types/counter.rb
```diff
@@ -1,20 +1,16 @@
...
+  before_method :set_default, :value, :increment, :decrement
+
   attr_accessor :expires_in
 
   def increment(by: 1)
-    multi do |pipeline|
-      pipeline.set 0, ex: expires_in, nx: true
-      pipeline.incrby by
-    end[-1]
+    incrby by
```
dhh, line 9: Isn't this a behavior change, then? Unless you manually specify the default to be zero?

lib/kredis/types/counter.rb
```diff
@@ -24,4 +20,17 @@ def value
   def reset
     del
   end
+
+  private
+
```
dhh, line 25: ✂️ No newline after `private`, just indention.

lib/kredis/types/counter.rb
```diff
@@ -24,4 +20,17 @@ def value
...
+    def value=(new_value)
+      set(new_value.to_i, ex: expires_in, nx: true)
+    end
+
+    def set_default
+      return if exists?
+
+      value = @default.is_a?(Proc) ? @default.call : @default
+      self.value = value
```
dhh, line 34: Why the local `value` variable? Can just assign directly. You don't need the guard clause either:

```ruby
self.value = resolve_default_value unless exists?
```

And then add `resolve_default_value`.

lib/kredis/types/flag.rb
```diff
@@ -8,10 +8,18 @@ def mark(expires_in: nil, force: true)
...
-    exists?
+    exists? || default?
   end
 
   def remove
     del
   end
+
+  private
+
```
dhh, line 19: ✂️

lib/kredis/types/flag.rb
```diff
@@ -8,10 +8,18 @@ def mark(expires_in: nil, force: true)
   end
 
   def marked?
-    exists?
+    exists? || default?
```
dhh, line 11: `exists? || default?` -> `exists? || exists_after_default_value?`

lib/kredis/types/flag.rb
```diff
@@ -8,10 +8,18 @@ def mark(expires_in: nil, force: true)
...
   end
 
   def remove
     del
   end
+
+  private
+
+    def default?
+      return !!@default unless @default.is_a?(Proc) && @default.call
```
dhh, line 21: Rename to `exists_after_default_value?` and then:

```ruby
def exists_after_default_value?
  if resolve_default_value == true
    mark
    true
  end
end
```

lib/kredis/types/hash.rb
```diff
@@ -42,4 +44,13 @@ def keys
   def values
     strings_to_types(hvals || [], typed)
   end
+
+  private
+
```
dhh, line 49: ✂️

lib/kredis/types/list.rb
```diff
@@ -24,4 +25,13 @@ def append(*elements, pipeline: nil)
   def clear
     del
   end
+
+  private
+
```
dhh, line 30: ✂️

lib/kredis/types/list.rb
```diff
@@ -24,4 +25,13 @@ def append(*elements, pipeline: nil)
...
+    def default
```
dhh, line 31: This has side effects, so needs a different name. Like `initialize_with_defaults`.

lib/kredis/types/proxy.rb
```diff
@@ -18,6 +19,16 @@ def multi(&block)
     end
   end
 
+  def get
+    super || default
+  end
+
+  def default
+    return @default unless @default.is_a? Proc
```
dhh, line 27: Drop the guard style and move to:

```ruby
if @default.is_a?(Proc)
  @default.call.tap { |value| set(value) unless value.nil? }
else
  @default
end
```

lib/kredis/types/proxying.rb
```diff
@@ -19,4 +23,10 @@ def failsafe(returning: nil, &block)
 
   private
     delegate :type_to_string, :string_to_type, :types_to_strings, :strings_to_types, to: :Kredis
+
+    def default
```
dhh, line 27: This is duplicated from the proxy object?

lib/kredis/types/scalar.rb
```diff
@@ -36,4 +30,12 @@ def expire_in(seconds)
   def expire_at(datetime)
     expireat datetime.to_i
   end
+
+  private
+
```
dhh, line 35: ✂️

lib/kredis/types/scalar.rb
```diff
@@ -36,4 +30,12 @@ def expire_in(seconds)
...
+    def default
```
dhh, line 36: Need to reflect side effects in the name like the other.

lib/kredis/types/set.rb
```diff
@@ -39,4 +40,13 @@ def take
   def clear
     del
   end
+
+  private
+
```
dhh, line 45: ✂️

lib/kredis/types/set.rb
```diff
@@ -39,4 +40,13 @@ def take
...
+    def default
```
dhh, line 46: Reflect side effect.

lib/kredis/types/counter.rb
```diff
@@ -24,4 +20,9 @@ def value
   def reset
     del
   end
+
+  private
+    def set_default
+      set(default.to_i, ex: expires_in, nx: true) unless exists?
```
dhh, line 26: Why do we need `unless exists?` if we're already using `nx: true`?

lib/kredis/types/flag.rb
```diff
@@ -8,10 +8,18 @@ def mark(expires_in: nil, force: true)
...
   def remove
     del
   end
+
+  private
+    def exists_after_default_value?
+      !!default do |default_value|
+        mark if default_value
+        !!default_value
+      end
```
dhh, line 23: This is very difficult to follow. Need to clarify this code somehow.

lib/kredis/types/counter.rb
tleish: Good catch.  Updated.

lib/kredis/types/flag.rb
tleish: Refactored

lib/kredis/types/set.rb
```diff
@@ -4,7 +4,8 @@ class Kredis::Types::Set < Kredis::Types::Proxying
   attr_accessor :typed
 
   def members
-    strings_to_types(smembers || [], typed).sort
+    value = exists? ? smembers : initialize_with_default || []
```
dhh, line 7: This adds an additional redis command call on every call to #members. Why wouldn't `smembers || initialize_with_default || []` work?

tleish: From my testing, when a _set_ has not been created, redis returns the following:

- `#exists?` => false
- `#smembers` => [] # empty array

I assumed we should not use just an empty array to decide if the code should apply the default value.

tleish: According to this answer on StackOverflow,

> When a set is empty, it is removed from the namespace.

see: https://stackoverflow.com/questions/13817865/redis-nil-or-empty-list-or-set#13820112

So, maybe we can assume with:

```ruby
strings_to_types(smembers.presence || initialize_with_default || [], typed).sort
```

dhh: That looks right to me 👍

tleish: Same question might go for list?

```diff
  def elements
-   value = exists? ? lrange(0, -1) : initialize_with_default || []
+   value = lrange(0, -1).presence || initialize_with_default || []
    strings_to_types(value, typed)
  end
```

dhh: Yes, same issue. Each redis command carries a cost over the network. Need to make sure we don't regress on that.

tleish: For hash and counter, this is going to be a challenge then.  

both of them implement a `before_methods` which checks if key exists before performing functions.

dhh: I think we have to find another way, then. We can't tax every hash and counter usage with an extra redis command just to support default.

tleish: We can solve this for counter, but not for Hash, List, Set, etc.  We can check if default is defined before checking `#exists?`, but I do not know of another way and open to suggestions.

dhh: Maybe @Jeremy has an idea.

tleish: Did a fairly significant refactor.  

- Used the same pattern throughout of `#initialize_with_default` defined in the base class.  `#initialize_with_default` calls `#set_default`, which is overridden in many of the child classes.
- Created a custom `#callnx` method which executes a redis lua script.  The lua script _calls_ a dynamic method with values if the redis key does not exist.  (Note: the script uses `redis#eval`, so additional care was taken to validate formatting of the method call.)  Using this custom redis method allows the code to run methods like hset and rpush only if key does not exist, and do this logic without adding redis command calls as it can be executed in a `multi` block
- Updated the custom redis `#multi` method to allow nested multi blocks (when one method using a multi block calls another method which also uses a multi block)

@dhh @jeremy - let me know if this a direction you want to take.

tleish: FYI, I tested performance with the following benchmarks


```ruby
require 'benchmark/ips'

proxy = Kredis.proxy 'mykey'
proxy.set 'myvalue'

Benchmark.ips(7) do |x|
  x.report("get only") do
    proxy = Kredis.proxy 'mykey'
    proxy.get
  end
  x.report("exists?") do
    proxy = Kredis.proxy 'mykey'
    proxy.set 'myvalue' unless proxy.exists?
    proxy.get
  end
  x.report("nx: true") do
    proxy = Kredis.proxy 'mykey'
    proxy.set 'myvalue', nx: true
    proxy.get
  end
  x.report("callnx") do
    proxy = Kredis.proxy 'mykey'
    proxy.callnx(:set, 'myvalue')
    proxy.get
  end
  x.report("multi > nx: true") do
    proxy = Kredis.proxy 'mykey'
    proxy.multi do
      proxy.set 'myvalue', nx: true
      proxy.get
    end
  end
  x.report("multi > callnx") do
    proxy = Kredis.proxy 'mykey'
    proxy.multi do
      proxy.callnx(:set, 'myvalue')
      proxy.get
    end
  end
  x.compare!
end
```

The results show:

```
Warming up --------------------------------------
            get only     1.064k i/100ms
             exists?   731.000  i/100ms
            nx: true   644.000  i/100ms
              callnx   621.000  i/100ms
    multi > nx: true   702.000  i/100ms
      multi > callnx   664.000  i/100ms
Calculating -------------------------------------
            get only     10.617k (± 6.7%) i/s -     74.480k in   7.053390s
             exists?      7.322k (± 2.9%) i/s -     51.901k in   7.094167s
            nx: true      7.030k (± 5.7%) i/s -     49.588k in   7.080403s
              callnx      6.810k (± 7.1%) i/s -     47.817k in   7.059580s
    multi > nx: true      6.891k (± 4.7%) i/s -     48.438k in   7.047004s
      multi > callnx      6.608k (± 4.2%) i/s -     46.480k in   7.046974s

Comparison:
            get only:    10617.3 i/s
             exists?:     7322.1 i/s - 1.45x  (± 0.00) slower
            nx: true:     7030.3 i/s - 1.51x  (± 0.00) slower
    multi > nx: true:     6891.2 i/s - 1.54x  (± 0.00) slower
              callnx:     6810.3 i/s - 1.56x  (± 0.00) slower
      multi > callnx:     6607.6 i/s - 1.61x  (± 0.00) slower
```

#### Observations:
- As expected, that get only is the fastest
- As expected, defining a default value adds cost in checking the existence and setting if not exist
- Perhaps something is wrong with the test setup, but in these scenarios using `multi` adds additional overhead vs not using `multi`.  I expected the opposite.  So, I setup a 2nd test around the usage of `multi`.

```ruby
require 'benchmark/ips'

proxy = Kredis.proxy 'mykey'
proxy.set 'myvalue'

n = 1
Benchmark.ips(7) do |x|
  x.report("1 cmd") do
    proxy = Kredis.proxy 'mykey'
    proxy.get
  end
  x.report("2 cmds") do
        proxy = Kredis.proxy 'mykey'
        proxy.set 'myvalue'
        proxy.get
      end
      x.report("3 cmds") do
        proxy = Kredis.proxy 'mykey'
        proxy.set 'myvalue'
        proxy.set 'myvalue'
        proxy.get
      end
      x.report("multi > 1 cmd") do
        proxy = Kredis.proxy 'mykey'
        proxy.multi do
          proxy.get
        end
      end
      x.report("multi > 2 cmds") do
        proxy = Kredis.proxy 'mykey'
        proxy.multi do
          proxy.set 'myvalue'
          proxy.get
        end
      end
      x.report("multi > 3 cmds") do
        proxy = Kredis.proxy 'mykey'
        proxy.multi do
          proxy.set 'myvalue'
          proxy.set 'myvalue'
          proxy.get
        end
      end
      x.compare!
    end
```

```
Warming up --------------------------------------
               1 cmd     1.059k i/100ms
              2 cmds   709.000  i/100ms
              3 cmds   552.000  i/100ms
       multi > 1 cmd   797.000  i/100ms
      multi > 2 cmds   730.000  i/100ms
      multi > 3 cmds   610.000  i/100ms
Calculating -------------------------------------
               1 cmd     10.290k (± 6.8%) i/s -     72.012k in   7.034624s
              2 cmds      7.160k (± 3.1%) i/s -     50.339k in   7.037131s
              3 cmds      5.289k (± 6.5%) i/s -     36.984k in   7.026510s
       multi > 1 cmd      7.937k (± 5.1%) i/s -     55.790k in   7.049686s
      multi > 2 cmds      6.976k (± 3.7%) i/s -     48.910k in   7.021257s
      multi > 3 cmds      6.317k (± 5.0%) i/s -     44.530k in   7.069582s

Comparison:
               1 cmd:    10289.5 i/s
       multi > 1 cmd:     7937.3 i/s - 1.30x  (± 0.00) slower
              2 cmds:     7160.2 i/s - 1.44x  (± 0.00) slower
      multi > 2 cmds:     6975.8 i/s - 1.48x  (± 0.00) slower
      multi > 3 cmds:     6316.9 i/s - 1.63x  (± 0.00) slower
              3 cmds:     5289.0 i/s - 1.95x  (± 0.00) slower
```

#### Observations:
- Using multi adds overhead
- Using multi is faster once you reach 3+ commands.  Using only 1 or 2 is slower
- In the latest updated version there's lots of usage of `multi` (which I assumed would be faster).  After seeing these benchmarks, it's going to slightly slow down the current performance.  For example:

```diff
def value
-     value_after_casting = string_to_type(get, typed)

-     if value_after_casting.nil?
-       default
-     else
-       value_after_casting
-     end
+    get_value = multi do
+         set type_to_string(value, typed), ex: expires_in, nx: true if default_value.present?
+         get
+    end[-1]
+    string_to_type(get_value, typed)
end
```

In the above example, even if default is nil the code will run the `get` inside of `multi`, which goes from 10289.5 i/s to 7937.3 i/s.

Based on these benchmarks, I suggest either going back to using `#exists?` (if code defines default).

lib/kredis/types/hash.rb
```diff
@@ -1,28 +1,37 @@
 require "active_support/core_ext/hash"
 
 class Kredis::Types::Hash < Kredis::Types::Proxying
-  proxying :hget, :hset, :hmget, :hdel, :hgetall, :hkeys, :hvals, :del, :exists?
+  ZERO_FIELDS_ADDED = 0
+
+  proxying :hset, :hdel, :hgetall, :del, :exists?, :multi, :callnx
```
tleish, line 6: I removed some of these proxying methods because they were no longer used in the class.  Do they need to be kept because of the public api others may have tied into?

lib/kredis/types/set.rb
tleish: I have updated the code to be more efficient in that it only uses `multi` in the event that a default value has been defined, otherwise it reverts back to the prior behavior.

At this point, the code is ready for another review.

lewispb:
I wonder if we could simplify this significantly.

Rather than check (and set) a default value on type's read and write operations, could we initialize the default value once upon type instance initialization?

Here's what I mean:

```rb
class Person
  include Kredis::Attributes

  kredis_set :names, default: [ "Lewis", "Jeremy" ]

  def id = 1
end
```
### Setting default value on each operation (this PR)

```rb
>> person = Person.new
>> person.names

>> person.names.add "David" # every operation must be aware of defaults
  Kredis Proxy (0.0ms)  CALLNX people:1:names ["sadd", "Lewis", "Jeremy"]
  Kredis Proxy (0.0ms)  SADD people:1:names ["David"]
=> 1

>> person.names.members
  Kredis Proxy (0.0ms)  CALLNX people:1:names ["sadd", "Lewis", "Jeremy"]
  Kredis Proxy (0.0ms)  SMEMBERS people:1:names
=> ["David", "Jeremy", "Lewis"]
```

### Setting default value on initialization

```rb
>> person = Person.new
>> person.names
  Kredis  (0.1ms)  Connected to shared
  Kredis Proxy (1.6ms)  EXISTS? people:1:names # One-time check...
  Kredis Proxy (0.2ms)  SADD people:1:names ["Lewis", "Jeremy"] # and set default

>> person.names.add "David" # subsequent operations don't need to concern themselves with defaults
  Kredis Proxy (0.7ms)  SADD people:1:names ["David"]
=> 1

>> person.names.members
  Kredis Proxy (0.6ms)  SMEMBERS people:1:names
=> ["David", "Jeremy", "Lewis"]
```

I think setting the default values on initialization has a few benefits:

- Simpler Ruby implementation, only type initialization needs to consider defaults rather than every operation
- No need for the Redis Lua script which adds overhead to the Redis server performance
- 2 additional Redis operations when a class is initialized, rather than N operations depending on usage

lewispb:
@tleish I hope you don't mind but I kept your tests but adjusted the implementation to fit my idea of setting defaults on initialization: https://github.com/rails/kredis/pull/119

PR #91, opened 2022-06-18
Extend expirable capability for any type

pedrofs:
Hello,

I need the expire feature for hash and saw some work done on expiring lists and unique lists. See #86. I thought about extending the expiration capability to any type. Perhaps it is valuable to others.

Here is a POC. I added the `Expirable` module to `List` and `Hash` to showcase.

When it has one, executing the expire under the `pipeline` is still missing (a problem we wouldn't need to deal with if we abstract the pipeline like #78 does).

Besides that, it could also make `expires_in` and `expires_at` inside `options` in the upstream API. Would it also make things easier?

Is it worth it to keep working here? @jorgemanrubia, do you mind reviewing it?

Thanks.

### DIFF
--- a/lib/kredis/types/expirable.rb
+++ b/lib/kredis/types/expirable.rb
@@ -1,0 +1,38 @@
+module Kredis::Types::Expirable
+  extend ActiveSupport::Concern
+
+  included do
+    proxying :expire, :expireat, :ttl
+
+    attr_accessor :expires_in, :expires_at
+
+    def expire_in(seconds)
+      expire seconds.to_i
+    end
+
+    def expire_at(datetime)
+      expireat datetime.to_i
+    end
+  end
+
+  def self.on(*on_methods)
+    Module.new do
+      define_singleton_method :included do |type_klass|
+        type_klass.include Kredis::Types::Expirable
+
+        type_klass.prepend(Module.new do
+          on_methods.each do |method|
+            define_method method do |*args|
+              super(*args)
+
+              expire_in(@expires_in) if expires_in
+              expire_at(@expires_at) if expires_at && !expires_in
+            end
+          end
+        end)
+      end
+    end
+  end
+end
+
+
--- a/lib/kredis/types/expirable.rb
+++ b/lib/kredis/types/expirable.rb
@@ -14,9 +14,13 @@ module Kredis::Types::Expirable
       expireat datetime.to_i
     end
 
+    def expire?
+      expires_in || expires_at
+    end
+
     def refresh_expiration
-      expire_in(@expires_in) if expires_in
-      expire_at(@expires_at) if expires_at && !expires_in
+      expire_in(expires_in) if expires_in
+      expire_at(expires_at) if expires_at && !expires_in
     end
   end
 
@@ -28,11 +32,14 @@ module Kredis::Types::Expirable
         type_klass.prepend(Module.new do
           on_methods.each do |method|
             define_method method do |*args, **kargs, &block|
-              operation_result = super(*args, **kargs, &block)
+              unless expire?
+                super(*args, **kargs, &block)
+              else
+                operation_result = super(*args, **kargs, &block)
+                refresh_expiration
 
-              refresh_expiration
-
-              operation_result
+                operation_result
+              end
             end
           end
         end)
--- a/lib/kredis/types/expirable.rb
+++ b/lib/kredis/types/expirable.rb
@@ -27,8 +27,8 @@ module Kredis::Types::Expirable
 
         type_klass.prepend(Module.new do
           on_methods.each do |method|
-            define_method method do |*args|
-              super(*args)
+            define_method method do |*args, **kargs, &block|
+              super(*args, **kargs, &block)
 
               refresh_expiration
             end
--- a/lib/kredis/types/expirable.rb
+++ b/lib/kredis/types/expirable.rb
@@ -13,6 +13,11 @@ module Kredis::Types::Expirable
     def expire_at(datetime)
       expireat datetime.to_i
     end
+
+    def refresh_expiration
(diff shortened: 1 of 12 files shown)


lib/kredis/types/expirable.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+
+    def expire_at(datetime)
+      expireat datetime.to_i
+    end
+
+    def refresh_expiration
+      expire_in(@expires_in) if expires_in
+      expire_at(@expires_at) if expires_at && !expires_in
+    end
```
swanson, line 20: You may consider adding a `clear_expiration` that does `expire 0` or `persist` Redis commands...but that could be a separate addition.

test/types/hash_test.rb
```diff
@@ -15,6 +15,19 @@ class HashTest < ActiveSupport::TestCase
     @hash[:key]  = :value
     @hash[:key2] = "value2"
     assert_equal({ "key" => "value", "key2" => "value2" }, @hash.to_h)
+    assert_equal(-1, @hash.ttl)
+  end
+
+  test "[]= assigment with expires_in" do
+    @hash.expires_in = 25
+    @hash[:key] = :value
+    assert(@hash.ttl.between?(20, 25))
```
swanson, line 24: In my testing, as long as we are calling the assertion within one second of setting the ttl, I was able to assert on the value directly. I ran the tests ~20 times without any flaky failures. The `ttl` only has one second precision.

swanson:
I think this looks good to me and seems appropriate since the behavior of expiring keys is not really specific to a particular type. :+1:

test/types/hash_test.rb
pedrofs: I ran it many times without an issue too. But it happens to fail.

lib/kredis/types/expirable.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def self.on(*on_methods)
+    Module.new do
+      define_singleton_method :included do |type_klass|
+        type_klass.include Kredis::Types::Expirable
+
+        type_klass.prepend(Module.new do
+          on_methods.each do |method|
+            define_method method do |*args, **kargs, &block|
+              super(*args, **kargs, &block)
```
jorgemanrubia, line 31: Notice this would change the return method. You could use `tap`:

```ruby
super(*args, **kargs, &block).tap do
  refresh_expiration
end
```

I initially thought that there should be a transaction wrapping the original code and the `refresh_expiration logic`. But then I thought that it's probably not worth the hassle. I think it's acceptable that these `expire` operations aren't executed transactionally with the intercepted operation. In case of a race condition, the last refresh would prevail, and that's fine. If we wanted to do it, the problem is that Redis doesn't support nested `multi` operations. If we got #78 merged, maybe we could use that to detect if there is an ongoing transaction or we need to start one (assuming there is an expiration date to use). But as mentioned, I don't think it's worth the trouble.

lib/kredis/types/expirable.rb
```diff
@@ -0,0 +1,42 @@
...
+  included do
+    proxying :expire, :expireat, :ttl
+
+    attr_accessor :expires_in, :expires_at
+
+    def expire_in(seconds)
+      expire seconds.to_i
+    end
+
+    def expire_at(datetime)
```
jorgemanrubia, line 13: If we kept `expire_at` we should review existing cases where we use `expire_in` to support this. E.g: add `:exat` support to `#set` operations using `:ex`. Maybe we can only address `expire_in` in the initial implementation?

I think consistency is important. So if we want to support `expire_at`, I'd make sure it's supported across all types.

jorgemanrubia (commented):
@pedrofs I love the feature and overall implementation approach!

Please, could you extend this to the rest of types and document in the `README`? Having a homogeneous `expire_in` across all attribute types is a wonderful addition to the library.

We'll get merged #78 too.

> Besides that, it could also make expires_in and expires_at inside options in the upstream API. Would it also make things easier?

I am not sure what do you have in mind there. Could you elaborate?

lib/kredis/types/expirable.rb
swanson: There is also an option when sending the EXPIRE command to Redis to only update if the new value is greater than the existing TTL. That could be helpful for the race-condition case.

lib/kredis/types/expirable.rb
pedrofs: I'm not a Redis expert, but what if we stopped using `ex` and relied solely on the proposed implementation?

jorgemanrubia: We'd lose the atomicity of doing both operations in a single one, but I think it's a fair tradeoff. We can always optimize things later if we find the need to do it.

I have mixed feelings about not doing the expire within a transaction. It can happen that the client loses the connection to redis right before setting the `expire`, letting the key without the expiration date set behind.

pedrofs: I agree. I'll start to implement it to all other types, and when #78 gets merged, I can give it a shot for the approach you suggested and start a transaction if we are not in yet. Is it worth it?

lib/kredis/types/expirable.rb
pedrofs: > Notice this would change the return method. You could use tap:

Not that easy, [`Redis::Future` inherits from `BasicObject`](https://www.rubydoc.info/github/redis/redis-rb/Redis/Future).

turgs:
:wave: Is this good to go @pedrofs @jorgemanrubia ? I'd love this

pedrofs:
Hey, it is not good to go.

I identified three cases that would require a more robust approach. I have a newborn and no time to work here at the moment. I'll close the pull request for now and will try to open it again in the future.

Thanks.

PR #93, opened 2022-07-01
Configure fail safe mechanism

lewispb:
WIP

Adds a fail safe configuration mechanism

### DIFF
--- a/lib/kredis/types/proxy/failsafe.rb
+++ b/lib/kredis/types/proxy/failsafe.rb
@@ -7,19 +7,10 @@ module Kredis::Types::Proxy::Failsafe
 
   def failsafe(returning: nil, &block)
     if fail_safe_enabled? && !fail_safe_suppressed?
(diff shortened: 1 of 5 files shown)


lib/kredis/types/proxy/failsafe.rb
```diff
@@ -1,16 +1,21 @@
...
-    @fail_safe_suppressed = false
+  extend ActiveSupport::Concern
+
+  included do
+    mattr_accessor :fail_safe_enabled, default: true
   end
 
-  def failsafe
-    yield
+  def failsafe(returning: nil, &block)
```
intrip, line 8: What about changing `failsafe(returning: nil, &block)` API to failsafe(returning: nil, errors: [Redis::BaseError], &block) such that we make explicit which errors we catch?

lewispb:
Closing in favor of https://github.com/rails/kredis/pull/94

PR #94, opened 2022-07-11
Permit disabling fail safe

lewispb:
Makes the fail safe mechanism optional. When fail safe is disabled, errors will be raised as per usual.

This covers scenarios where the global fail safe is undesirable.

**Usage**

```ruby
config.kredis.fail_safe = false
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,7 +249,7 @@ You can configure how the redis client is created by setting `config.kredis.conn
 config.kredis.connector = ->(config) { SomeRedisProxy.new(config) }
 ```
 
-The fail safe mechanism supports silently rescuing or returning a default value in the event that the Redis client returns an error (e.g. Redis is down). You can disable the default fail safe mechanism:
+The fail-safe mechanism supports silently rescuing or returning a default value in the event that the Redis client returns an error (e.g. Redis is down). You can disable the default fail-safe mechanism:
 
 ```ruby
 config.kredis.fail_safe = false
--- a/README.md
+++ b/README.md
@@ -241,12 +241,18 @@ Kredis::Connections.connections[:shared] = Redis.new(
 
 The above code could be added to either `config/environments/production.rb` or an initializer. Please ensure that your client private key, if used, is stored your credentials file or another secure location.
 
-### Configure how the redis client is created
+### Configuration
 
 You can configure how the redis client is created by setting `config.kredis.connector` in your `application.rb`:
 
 ```ruby
 config.kredis.connector = ->(config) { SomeRedisProxy.new(config) }
+```
+
+The fail safe mechanism supports silently rescuing or returning a default value in the event that the Redis client returns an error (e.g. Redis is down). You can disable the default fail safe mechanism:
+
+```ruby
+config.kredis.fail_safe = false
 ```
 
 By default Kredis will use `Redis.new(config)`.
--- a/lib/kredis/types/proxy/failsafe.rb
+++ b/lib/kredis/types/proxy/failsafe.rb
@@ -1,12 +1,10 @@
 module Kredis::Types::Proxy::Failsafe
   extend ActiveSupport::Concern
 
-  included do
-    mattr_accessor :fail_safe_enabled, default: true
-  end
+  mattr_accessor :enabled, default: true
 
   def failsafe(returning: nil, &block)
-    if fail_safe_enabled? && !fail_safe_suppressed?
+    if enabled? && !suppressed?
       rescue_redis_errors_with(returning: returning, &block)
     else
       yield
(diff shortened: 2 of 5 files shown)


test/proxy_test.rb
```diff
@@ -19,4 +19,14 @@ class ProxyTest < ActiveSupport::TestCase
...
   end
+
+  test "with fail safe disabled an error is raised" do
+    @proxy.fail_safe_enabled = false
+
+    stub_redis_down(@proxy) do
+      assert_raises(Redis::BaseError) { @proxy.get }
+    end
+
+    @proxy.fail_safe_enabled = true
```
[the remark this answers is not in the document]
lewispb: I don't think we can assume that `@proxy` will contain an instance of a `Proxy` (see migration_test.rb).

I do agree we should teardown though, so I added the teardown to the specific test instead.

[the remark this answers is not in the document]
jorgemanrubia: Yes, totally, I didn't notice extracting to the parent didn't make sense with the instance var 😅.

jorgemanrubia (approved):

README.md
```diff
@@ -241,14 +241,20 @@ Kredis::Connections.connections[:shared] = Redis.new(
...
 ```ruby
 config.kredis.connector = ->(config) { SomeRedisProxy.new(config) }
 ```
 
+The fail safe mechanism supports silently rescuing or returning a default value in the event that the Redis client returns an error (e.g. Redis is down). You can disable the default fail safe mechanism:
+
+```ruby
+config.kredis.fail_safe = false
+```
+
```
jeremy, line 257: Small thing, would use "failsafe" for noun usage and hyphenated "fail-safe" for descriptive adjectives.

lib/kredis/types/proxy/failsafe.rb
```diff
@@ -1,26 +1,33 @@
...
-    @fail_safe_suppressed = false
-  end
+  extend ActiveSupport::Concern
 
-  def failsafe
-    yield
-  rescue Redis::BaseError
-    raise if fail_safe_suppressed?
+  included do
+    mattr_accessor :fail_safe_enabled, default: true
```
jeremy, line 5: This attr can live on Kredis::Types::Proxy::Failsafe itself without mixing in to the including class and, in this context, it can be shorted to `:enabled`

README.md
lewispb: Thanks, updated. Also checked https://www.collinsdictionary.com/dictionary/english/fail-safe.

lib/kredis/types/proxy/failsafe.rb
lewispb: Good suggestion. I also changed fail_safe_suppressed -> suppressed.

jeremy (commented):
Continuing reflection on the role of a proxy/connection-level fail-safe…

At the center of this is the blanket fail-safe in `Kredis::Types::Proxy#method_missing`. That applies to all proxied methods, which means all Redis commands that each type invokes.

Do all our data structures appropriately anticipate that the Redis commands they invoke could silently do nothing? On an initial perusal, they do not, and the consequences range from incorrect behavior to data corruption.

Furthermore, hiding errors means that other error handlers (like circuit breakers and exception reporters) never have the chance to see that something's going wrong.

I think it's worth upping the scrutiny here and probably removing the current failure-safety implementation.

Fail-safe behavior should generally be a characteristic of the data types, not the Redis connection, and it needs to be consistently applied. `Slots#reserve` and `#available?` do this, for example, yet `Slots#reset` does not.

Furthermore, in the context of data types (rather than the proxy/Redis connection) this looks more like an "offline mode" than low-level failure-safety, which suggests a simple design path: when fail-open/closed-when-offline behavior is warranted, implement it by composing or subclassing the basic Kredis type and wrapping it with appropriate exception handling.

So, for the purposes of this PR, I think we should park it 🥺 and not introduce new global config since there really isn't coherent, toggleable failure-safety behavior to expose. Doing so would ultimately complicate efforts to rectify this situation since we'd need to ditch the config afterward.

lewispb:
Great assessment, thank you @jeremy. You're right. Let's close this for now and instead move to rethink the fail-safe behaviour.

maxim:
@jeremy I agree as well that most error handling of this kind is ideally decided by the caller, but on the other side, most people just don't want their apps to crash because, say, a limiter is not working.

That said, I think most people would want to find out quickly that Redis is not working, for which normally they'd use some logger or notifier. We probably don't want to encourage subclassing and overriding the failsafe every time someone wants to add a notifier. It would make future changes/refactorings of kredis likely to break codebases.

Which leads me to think that maybe the right approach is for some types to accept an optional callback that passes in the error. In the callback we can send a notice. This gives people a stable api, allowing internals to change freely.

PR #96, opened 2022-08-22
Add `LTRIM` to lists

etiennebarrie:
See https://redis.io/commands/ltrim/


Schwad:
+1 - this would serve us well in a current migration to Kredis that I'm working!

PR #97, opened 2022-08-22
Add `last` to lists

etiennebarrie:
This adds support for `last` for lists. Both `last` and `last(n)` are supported.

Because Ruby doesn't provide a great way to replicate the implicit conversions (`Integer.try_convert` exists in Ruby 3.1, but we'd still need to generate the error message manually because it returns `nil` instead of raising the `TypeError`), some edge cases of the implementation of `Array#last` are not replicated:
- `[].last(nil)` raises a `TypeError`, whereas we just fail with <code>undefined method `to_int' for nil:NilClass (NoMethodError)</code>
- `[].last(object)` where `object.to_int` doesn't return an `Integer` raises a `TypeError`, whereas we will call `==` and `<` on `object`, and the results may be unexpected (though that is a really edgy case).

The first commit is really the most straightforward implementation, and we may not need the second one if we don't care about the edge case of `nil` or `false` being passed and expecting an exception. I think in most cases `last(n)` is called with a literal value, so it's fine.

### DIFF
--- a/lib/kredis/types/list.rb
+++ b/lib/kredis/types/list.rb
@@ -26,16 +26,6 @@ class Kredis::Types::List < Kredis::Types::Proxying
   end
 
   def last(n = nil)
-    if n
-      if n == 0
-        []
-      elsif n < 0
-        raise ArgumentError, "negative array size"
-      else
-        lrange(-n, -1)
-      end
-    else
-      lrange(-1, -1).first
-    end
+    n ? lrange(-n, -1) : lrange(-1, -1).first
   end
 end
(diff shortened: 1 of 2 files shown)


rafaelfranca:
I find this implementation already too complex. This should not be receiving user input, so I don't think we should be even calling `to_int`. Just assume arguments are going to be integers already and let's handle `nil`. All rest should just assume we are getting input from developers and if they want to pass something that isn't an integer it should just not work.

etiennebarrie:
I removed the `to_int` and assumed `nil` is equivalent to no argument being passed.

PR #110, opened 2023-04-25
adds hincrby and hincrbyfloat

JasonMannon:
Adds support for [HINCRBY](https://redis.io/commands/hincrby/) and [HINCRBYFLOAT](https://redis.io/commands/hincrbyfloat/)


dhh:
We're not trying to expose all the underlying redis functions directly. Only as they can be exposed in the context of Ruby structures. You can always access these commands directly via Redis.

PR #119, opened 2023-07-08
Default values on initialization

### LINKED ISSUE #66: Default proc if empty

lewispb:
I built upon what was started in https://github.com/rails/kredis/pull/89 but attempted a simpler implementation, only setting default values upon type initialization.

Closes #66 and #89

Almost all the same tests pass, with one or two minor differences.

**Todo**
- [x] Update readme
- [x] Code style (see https://github.com/rails/kredis/pull/120)


test/types/enum_test.rb
```diff
@@ -37,9 +49,6 @@ class EnumTest < ActiveSupport::TestCase
   end
 
   test "exists?" do
-    assert_not @enum.exists?
-
-    @enum.value = "one"
```
lewispb, line 42: An enum always has a default and so always now exists.

dhh:
I like this a lot. Nice simplification 👍

jeremy (commented):
Enum is a bit of an oddball re. defaults out of the enumerated set and `nil` defaults in particular. From HEY:
```
kredis_enum :status, values: %w[ waiting confirmed ], default: nil, key: …
```
Defaults out of the enumerated set should probably be considered a bug, with a special case for `default: nil` behaving how an `optional: true` argument would.

lib/kredis/default_values.rb
```diff
@@ -0,0 +1,28 @@
...
+      def set_default
+        raise NotImplementedError, "Kredis type #{self.class} needs to define #set_default"
+      end
+  end
+
+  def initialize(...)
+    super
+
+    if default.present? && !exists?
+      set_default
```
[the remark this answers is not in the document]
lewispb: Great point. Not possible with MULTI, can't check-and-set. Implemented with WATCH, and tested.

lib/kredis/attributes.rb
```diff
@@ -90,6 +90,7 @@ def kredis_connection_with(method, name, key, **options)
           if instance_variable_defined?(ivar_symbol)
             instance_variable_get(ivar_symbol)
           else
+            options[:default] = kredis_default_evaluated(options[:default]) if options[:default]
             new_type = Kredis.send(type, kredis_key_evaluated(key) || kredis_key_for_attribute(name), **options)
```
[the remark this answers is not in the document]
lewispb: cycle, slot and slots types don't support a default kwarg, so can't blindly send that kwarg on to all types if it's not supplied.

lib/kredis/types/enum.rb
```diff
@@ -33,4 +38,8 @@ def define_predicates_for_values
         define_singleton_method("#{defined_value}!") { self.value = defined_value }
       end
     end
+
+    def set_default
+      self.value = default
+    end
```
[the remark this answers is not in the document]
lewispb: Fixed!

lewispb:
> Defaults out of the enumerated set should probably be considered a bug, with a special case for nil

Thanks @jeremy, agree, and updated to raise an error in that case.

jeremy (approved):

README.md
```diff
@@ -197,6 +189,29 @@ person.morning.value = "blue"                        # => SET people:5:morning
...
+  kredis_list :favorite_colors, default: [ "red", "green", "blue" ]
+end
+```
+
+There's a performance overhead to consider though. When you first read or write an attribute in a model, Kredis will
+check if the underlying Redis key exists, while watching for concurrent changes, and if it does not,
+write the specified default value.
+
+This means that using default values in a typical Rails app additional Redis calls (WATCH, EXISTS, UNWATCH) will be
+executed for each Kredis attribute with a default value read or written during a request.
```
tleish, line 211: Does this occur every request, or only when a value does not exist?

lewispb: In the context of a Rails web request, as the Kredis object is initialized, the key is checked with WATCH, EXISTS, UNWATCH once, but then subsequent operations within the request do not incur any Redis command overhead. If no default value is specified, no additional Redis command overhead is incurred at all.

lewispb: ### Scenario 1: Default NOT defined
(no additional overhead)

### Scenario 2: Default defined, value NOT exists
WATCH, EXISTS, write default (depending on type), UNWATCH

### Scenario 3: Default defined, value exists
WATCH, EXISTS, UNWATCH

tleish: It seems a more efficient approach would be to only add extra overhead of default is defined AND value doesn't exist. In other words:

1. Get Value
2. Value doesn't exist
3. Execute Default Logic

This avoids extra overhead if the value has been set already.

lewispb: I agree @tleish. That's exactly scenario 3, but with the added WATCH / UNWATCH.

To illustrate the problem that the WATCH / UNWATCH solves check this test case:

https://github.com/rails/kredis/blob/b1a818ad3e70a23154ed90105a6f1d6f060cbbd1/test/types/counter_test.rb#L113-L121

Without the WATCH / UNWATCH, the value existing check is prone to a race condition.

nzwsch:
I noticed that the default value cannot be set in kredis_integer in the current release version 1.5.0.
```
Failure/Error: kredis_integer :foo, default: 0

ArgumentError:
  unknown keyword: :default
```
When will kredis be released with this PR?

dhh:
1.6.0 released 👍

PR #120, opened 2023-07-08
Set up RuboCop with CI

lewispb:
Brings in RuboCop with the same configuration (and gem dependencies) as https://github.com/rails/rails

Also adds a CI step to lint PRs, again as per https://github.com/rails/rails.

There are 76 offences currently, I can address those in a subsequent PR.


jeremy (commented):
Would include the fixes here in a second commit; needn't leave CI red.

guilleiguaran (approved):

PR #122, opened 2023-07-09
Support line filtering in tests

lewispb:
Line filtering was not working when running tests with `bin/test`.

e.g. `bin/test test/attributes_test.rb:82` would run all tests, not only the one defined on line 82.

This is because `rails/plugin/test` does not support line filtering. We should probably fix that upstream.

I opened https://github.com/rails/rails/issues/48696 to track that.

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -1,6 +1,7 @@
 require "bundler/setup"
 require "active_support/test_case"
 require "active_support/testing/autorun"
+require "rails/test_unit/line_filtering"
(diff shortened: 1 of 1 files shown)


yahonda:
It works as expected.

```
$ bin/test test/attributes_test.rb:82
Run options: --seed 43010

# Running:

.

Finished in 0.011638s, 85.9265 runs/s, 85.9265 assertions/s.
1 runs, 1 assertions, 0 failures, 0 errors, 0 skips
$
```

Issue #116, opened 2023-06-27
Is there any reason about Kredis.boolean set redis value to "true", "false" instead of "1", "0" ?


roharon:
I wonder about why `Kredis.boolean` set value to `"true"` or `"false"`, not `"1"`, `"0"` 

```ruby
boolean = Kredis.boolean "myboolean"

boolean.value = true
##
## below this line is work In redis 
# get myboolean
# "true"

boolean.value = false
##
## below this line is work In redis 
# get myboolean
# "false"
```

I think set boolean value to `"true"` or `"false"` could be inefficient rather than `"0"`, `"1"` from the standpoint of Redis memory usage.

```shell
> set myboolean true
OK
> MEMORY USAGE myboolean
(integer) 72

> set myboolean false
OK
> MEMORY USAGE myboolean
(integer) 72

> set myboolean 1
OK
> MEMORY USAGE myboolean
(integer) 56

> set myboolean 0
OK
> MEMORY USAGE myboolean
(integer) 56

```

Also, I checked when get "myboolean" key's value likes below code, it casted to boolean by `ActiveModel::Type::Boolean`
```ruby
boolean = Kredis.boolean "myboolean"

## In redis
# SET myboolean 0
boolean.value
# active_model/type/boolean.rb cast it to `false` 


## In redis
# SET myboolean 1
boolean.value
# active_model/type/boolean.rb cast it to `true`
``` 

So I want to suggest about `Kredis.boolean` set redis value to "1", "0".

lewispb:
Agree. Opened https://github.com/rails/kredis/pull/123 @roharon

PR #123, opened 2023-07-09
Make Boolean storage more efficient

lewispb:
Closes https://github.com/rails/kredis/issues/116

Rather than store the boolean as a string, store it as a 1 or 0. This is more efficient in terms of storage.

This change is backwards-compatible, so existing boolean values will continue to work as expected.

cc @roharon

### DIFF
--- a/test/types/scalar_test.rb
+++ b/test/types/scalar_test.rb
@@ -38,6 +38,22 @@ class ScalarTest < ActiveSupport::TestCase
     assert_equal false, boolean.value
   end
 
+  test "boolean casting" do
+    boolean = Kredis.boolean "myscalar"
+
+    boolean.value = true
+    assert_equal "1", boolean.get
+
+    boolean.value = false
+    assert_equal "0", boolean.get
+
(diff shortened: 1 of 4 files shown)


test/types/scalar_test.rb
```diff
@@ -38,6 +38,22 @@ class ScalarTest < ActiveSupport::TestCase
...
+  test "boolean casting" do
+    boolean = Kredis.boolean "myscalar"
+
+    boolean.value = true
+    assert_equal "1", boolean.get
+
+    boolean.value = false
+    assert_equal "0", boolean.get
+
+    boolean.set "true"
```
roharon, line 50: Is it neccessary on this test case?

lewispb: I’m testing backwards compatibility here so I think it’s necessary.

roharon: I see

jeremy (approved):

guilleiguaran (approved):

PR #126, opened 2023-08-21
Adds include? to ordered set

EduardoAntequera:
Given that ordered set is using the underlying Redis sorted set, I believe it can be usefull to have the 'include?' function also there to check if an element or number of elements are already there by using Redis zrank.

### DIFF
--- a/lib/kredis/types/ordered_set.rb
+++ b/lib/kredis/types/ordered_set.rb
@@ -3,7 +3,7 @@
 class Kredis::Types::OrderedSet < Kredis::Types::Proxying
   prepend Kredis::DefaultValues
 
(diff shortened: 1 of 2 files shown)


lib/kredis/types/ordered_set.rb
```diff
@@ -17,6 +17,10 @@ def remove(*elements)
     zrem(types_to_strings(elements, typed))
   end
 
+  def include?(element)
+    zrank(type_to_string(element, typed)).present?
```
lewispb, line 21: I'd suggest using zscore instead which is O(1) instead of O(log(N)). I'd also avoid `.present?` in case an empty string is a valid set member.

```suggestion
    !!zscore(type_to_string(element, typed))
```

EduardoAntequera: Good points :+1:

PR #132, opened 2023-11-06
Fix rubocop warnings

nzwsch:
I recently made a comment urging the release of kredis.
Thank you for releasing 1.6.0.
I don't know if this is a result of that, but I noticed that the tests in the main branch are red.
Fortunately, I ran Rubocop locally and all errors were resolved.
I don't know if this qualifies as an open source contribution, but let me submit it as a PR.

Here is my output:

```
$ bundle exec rubocop
Inspecting 56 files
........................................................

56 files inspected, no offenses detected
```

### DIFF
--- a/lib/kredis/version.rb
+++ b/lib/kredis/version.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module Kredis
   VERSION = "1.6.0"
 end
--- a/test/attributes_test.rb
+++ b/test/attributes_test.rb
@@ -160,7 +160,7 @@ class AttributesTest < ActiveSupport::TestCase
     assert_not @person.special?
 
     @person.special.mark
-    assert @person.special?
+    assert_predicate @person, :special?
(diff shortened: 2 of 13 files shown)


dhh:
I don't like any of those fixes. Should tell rubocop to ignore them instead.

nzwsch:
We have two options.

The first is that this Rubocop rule references a rule in Rails itself, and Rails allows `AssertPredicate`, so we need to [disable this one](https://github.com/rails/rails/blob/main/.rubocop.yml#L358-L359).
The second would be to add `AssertPredicate` to the .rubocop.yml file as an exception.

Which is preferred?

dhh:
Let's adopt https://github.com/rails/rubocop-rails-omakase/.

dhh:
Fixed in https://github.com/rails/kredis/commit/01a893b267f3456a6eb18f761b2b5425654b8c1c.

PR #141, opened 2024-01-09
Redundant limiter.poke in README.md

ermolaev:
```ruby
limiter = Kredis.limiter "mylimit", limit: 3, expires_in: 5.seconds
7.times { pp [limiter.poke, limiter.exceeded?] }
[1, false]
[2, false]
[3, true]
[4, true]
[5, true]
[6, true]
[7, true]
```


jeremy (approved):
Indeed, it's `>=` not `>`.

PR #142, opened 2024-01-26
Support `expires_in` for list

heka1024:
Closes https://github.com/rails/kredis/issues/129

Started from https://github.com/rails/kredis/pull/133, I made a small change to support expiration time for lower versions of Redis.


myxoh (commented):
Any arguments against making this available in all datatypes where redis supports this?

jeremy:
#143

PR #143, opened 2024-02-01
Support `expires_in` for `List`, `Set`, `UniqueList`, `OrderedSet`

heka1024:
Extending the approach taken by https://github.com/rails/kredis/pull/142, now users can set `expires_in` for `List`, `Set`, `UniqueList`, `OrderedSet`

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,18 @@ limiter.poke                    # => SET limiter 0 NX + INCRBY limiter 1
 false == limiter.exceeded?      # => GET "limiter"
 ```
 
+Lists, unique lists, sets, and ordered sets support expiration:
+
+```ruby
+set = Kredis.set "myset", expires_in: 1.second
+set.add "hello", "world" # => SADD myset "hello" "world"
+true == set.include?("hello") # => SISMEMBER myset "hello
+sleep 2
+[] == set.members # => SMEMBERS myset
+```
+
+To support lower versions of redis, which does not has `nx` option on `EXPIRE` command, multiple commands are used to achieve the same effect.
+
 ### Models
 
 You can use all these structures in models:
@@ -193,6 +205,7 @@ class Person < ApplicationRecord
   kredis_unique_list :skills, limit: 2
   kredis_enum :morning, values: %w[ bright blue black ], default: "bright"
   kredis_counter :steps, expires_in: 1.hour
+  kredis_set :favorite_colors, expires_in: 1.day
 
   private
     def generate_names_key
--- a/README.md
+++ b/README.md
@@ -181,7 +181,7 @@ limiter.poke                    # => SET limiter 0 NX + INCRBY limiter 1
 false == limiter.exceeded?      # => GET "limiter"
 ```
 
-Lists, unique lists, sets, and ordered sets support expiration:
+Lists, unique lists, sets, ordered sets, and hashes support expiration:
 
 ```ruby
 set = Kredis.set "myset", expires_in: 1.second
--- a/lib/kredis/types/hash.rb
+++ b/lib/kredis/types/hash.rb
@@ -4,6 +4,7 @@ require "active_support/core_ext/hash"
 
 class Kredis::Types::Hash < Kredis::Types::Proxying
   prepend Kredis::DefaultValues
+  include Kredis::Expiration
 
   proxying :hget, :hset, :hmget, :hdel, :hgetall, :hkeys, :hvals, :del, :exists?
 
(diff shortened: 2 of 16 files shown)


bb:
Hi @heka1024, thanks for this PR. I was looking for expiration on sets and happy to find your work.

Do you plan to also add expiration to the ActiveModel attributes via the [Kredis::Attributes](https://github.com/rails/kredis/blob/main/lib/kredis/attributes.rb) concern?

Also, I'm afraid to get this PR merged, some documentation is needed.

heka1024:
@bb Thanks. I'll chery pick your commit and supplement documentation.

heka1024:
@bb I just cherry-picked your commit and add some test with documentation. Could you review it?

bb:
Hey @heka1024, thanks a lot!

The tests you already added in the last commits look good to me. I'm afraid there's no test for ordered set yet.

Docs look also good, I don't think exhaustive examples of all params is needed here, so I'd leave it as is. Keep in mind I'm not a maintainer here, so in the end it's not my decision.

bb:
Thanks for adding the missing test @heka1024. LGTM.
As said, I'm no maintainer, so I cannot formally approve it.

Maybe @dhh (or another maintainer) can have a look now because his concern about test failures in https://github.com/rails/kredis/pull/133#issuecomment-1872184287 is addressed now.

balbesina:
any reasons why this list doesn't include Hash ?

heka1024:
@balbesina. I just added expiration behavior in hash. I rarely used Redis's hash, I just forgot it 😅.

heka1024:
Hi, @dhh. Could you review this? I fixed https://github.com/rails/kredis/pull/133#issuecomment-1872184287's failing tests.

roharon:
Although this is a different topic, have you considered adding an expires_at option?
I believe it would be beneficial for [product or feature] to also support expires_at. If it hasn't been considered yet, I'm thinking about adding expires_at and would like to follow up on it.

heka1024:
@roharon It'll be nice if we can set `expires_at`! Could you create PR for it?

roharon:
@heka1024  Cool, I'll work on it.

thiagopradi:
@heka1024  - Please let me know if you need any help with this PR to get it merged. It will be useful for me too.

chaadow:
Please could this be merged? 🙏🏼

README.md
```diff
@@ -177,6 +177,18 @@ limiter.poke                    # => SET limiter 0 NX + INCRBY limiter 1
...
+
+```ruby
+set = Kredis.set "myset", expires_in: 1.second
+set.add "hello", "world" # => SADD myset "hello" "world"
+true == set.include?("hello") # => SISMEMBER myset "hello
+sleep 2
+[] == set.members # => SMEMBERS myset
+```
+
+To support lower versions of redis, which does not has `nx` option on `EXPIRE` command, multiple commands are used to achieve the same effect.
```
jeremy, line 190:
```suggestion
To support lower versions of Redis which do not have the `nx` option on `EXPIRE`, multiple commands are used to achieve the same effect.
```

lib/kredis/types/hash.rb
```diff
@@ -18,7 +19,9 @@ def []=(key, value)
   end
 
   def update(**entries)
-    hset entries.transform_values { |val| type_to_string(val, typed) }.compact if entries.flatten.any?
+    with_expiration do
```
jeremy, line 22: Refreshing expiration on modification differs from mainline Redis behavior.

If we want this, it should probably be distinct from setting an expiry on the key.

lib/kredis/types/list.rb
```diff
@@ -16,12 +17,20 @@ def remove(*elements)
     types_to_strings(elements, typed).each { |element| lrem 0, element }
```
jeremy, line 17: Removal doesn't freshen expiry, though?

Feeling like there will be lots of gaps in this behavior. Should be moved to a declaration on the underlying proxied operations?

lib/kredis/types/ordered_set.rb
```diff
@@ -53,9 +54,11 @@ def insert(elements, prepending: false)
...
       end
 
-      multi do
-        zadd(elements_with_scores)
-        trim(from_beginning: prepending)
+      with_expiration do
+        multi do
+          zadd(elements_with_scores)
+          trim(from_beginning: prepending)
+        end
```
jeremy, line 61: Similar to above, we have insertion refreshing expiration but not removal

jeremy (commented):
Would be clearer to introduce straightforward key expiration separately from refresh-on-change behavior. There seem to be a lot of gaps in that as well, suggesting a design rethink is in order.

heka1024:
Thank you for review, @jeremy . I'll take a look at it.

DavertMik:
This would be so nice to have!

PR #146, opened 2024-02-13
Kredis.hash fails with nil values in Redis 5+

swanson:
We've been bite several times in production by trying to set a kredis hash with the value of a key being `nil`. As we know, accidental `nil`s are not rare in Ruby :(

Kredis attempts to call `hset` in redis with a nil value, which is not supported and you get this exception:

```ruby
/Users/matt/.asdf/installs/ruby/3.3.0/lib/ruby/gems/3.3.0/gems/redis-client-0.19.1/lib/redis_client/command_builder.rb:37:in `block in generate': 
Unsupported command argument type: nil (TypeError)

            raise TypeError, "Unsupported command argument type: #{element.class}"
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

This PR runs `Hash#compact` on the values passed to Kredis before trying to send them to Redis via `hset`. This would remove invalid key/value pairs. Later, when reading the stored hash back, you would rely on the Ruby default return value of `nil` when accessing a non-existing key.

### DIFF
--- a/lib/kredis/types/hash.rb
+++ b/lib/kredis/types/hash.rb
@@ -18,7 +18,7 @@ class Kredis::Types::Hash < Kredis::Types::Proxying
   end
 
   def update(**entries)
-    hset entries.transform_values { |val| type_to_string(val, typed) } if entries.flatten.any?
+    hset entries.transform_values { |val| type_to_string(val, typed) }.compact if entries.flatten.any?
   end
 
   def values_at(*keys)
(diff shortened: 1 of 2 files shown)


swanson:
The root cause can be traced back to adding support for v5 `redis` gem: 

https://github.com/redis/redis-rb/issues/1142#issuecomment-1279910777

which changed the behavior (for good reason): https://github.com/redis/redis-rb/issues/1142#issuecomment-1288509586

jeremy:
If this is disallowed, seems we should raise ArgumentError on nil values rather than mask the issue by compacting.

And perhaps handled more broadly in `Kredis::TypeCasting` - does this affect hashes only?

PR #148, opened 2024-03-08
Add `include?` to list, uniquelist

heka1024:
# Motivation
- Support `include?` method for `Kredis::Types::List`

# Modification
- Allow `List` to proxy `lpos` (refer: https://redis.io/commands/lpos/)
- Add some example of `List#include?`

# Result
- Users can use `include?` without load members of list to memory


heka1024:
`lpos` requires version >= 6.0.6. Close PR until found a nice way to support lower versions.

jeremy:
Docs are sufficient here since it's a new capability, e.g. `list.include?` requires Redis 6+ with LPOS support. People can keep using `list.elements.include?` otherwise.

PR #149, opened 2024-03-08
Add `Kredis#redis_version`

heka1024:
There are many commands supported on only high versions of redis. For example `LPOS` or `NX` option in `EXPIRE`. So, if we can get version of current redis, we can implement some commands more effectively.

### DIFF
--- a/lib/kredis/info.rb
+++ b/lib/kredis/info.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+module Kredis::Info
+  def redis_version
+    redis_versions.first
+  end
+
+  def redis_versions
+    Array.wrap(Kredis.redis.info("server")).tap do |versions|
+      versions.map! { |v| v["redis_version"] }
+      versions.map! { |v| Gem::Version.new(v) }
+    end
+  end
+end
--- a/lib/kredis.rb
+++ b/lib/kredis.rb
@@ -13,11 +13,12 @@ require "kredis/type_casting"
 require "kredis/default_values"
 require "kredis/types"
 require "kredis/attributes"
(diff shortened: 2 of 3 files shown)


lib/kredis/info.rb
```diff
@@ -0,0 +1,14 @@
...
+  def redis_version
+    redis_versions.first
+  end
+
+  def redis_versions
+    Array.wrap(Kredis.redis.info("server")).tap do |versions|
+      versions.map! { |v| v["redis_version"] }
+      versions.map! { |v| Gem::Version.new(v) }
+    end
+  end
```
jeremy, line 13:
```suggestion
  def redis_version
    Gem::Version.new Kredis.redis.info("server")["redis_version"]
  end
```

jeremy (commented):
Too costly to use for runtime feature detection. Need a different approach if you'd like to use newer Redis capabilities.

Simplest approach, really, is to bump the minimum Redis version to something more modern. Redis 5 is ancient now.

heka1024:
@jeremy 
At first, I looked at rails' implementation and thought it wouldn't be a huge burnden. (https://github.com/rails/rails/blob/4cfdcfef8eedb28a4e0f5d12fae80c2aade02a4c/activesupport/lib/active_support/cache/redis_cache_store.rb#L450-L461) But as you said, it might cause some performance issue. Then, how about memoize result of `INFO` or define_method dynamically? What I thought is like https://github.com/rails/rails/blob/e5124aed3fdcacb05204391e236c4fd60a1e6e74/activesupport/lib/active_support/notifications/instrumenter.rb#L218-L226.

Issue #152, opened 2024-05-15
Kredis namespace isn't set properly for rails system tests


arjun810:
I ran into a bug where my system tests weren't running as expected when parallelized. I discovered that the Puma threads spawned by Capybara don't have the namespace set (see https://github.com/rails/kredis/blob/57cedf59a50a0ef81779b5259806821fc4a15423/lib/ kredis/railtie.rb#L8). This is because the namespacing is done using `Thread.current`, and the threads spawned by Puma wouldn't have `Thread.current[:kredis_namespace]` set.

I have a workaround for my system tests, but it'd be nice if this worked out of the box.

For my workaround, I have a module that my ApplicationSystemTestCase includes:

```
module KredisNamespaceFix
  def self.included(base)
    base.class_eval do
      setup do
        add_kredis_namespace_before_action
      end

      parallelize_setup do |worker|
        Rails.application.config.x.kredis_test_namespace = "test-#{worker}"
      end
    end
  end

  def add_kredis_namespace_before_action
    ApplicationController.class_eval do
      before_action :set_kredis_namespace_for_test

      private

      def set_kredis_namespace_for_test
        Kredis.namespace = Rails.application.config.x.kredis_test_namespace
      end
    end
  end
end
```

This seems to work, but it was a bit painful to track down.

jeremy:
That is painful. Ran into this also; partially fixed in #156.

Namespacing should be up-front config rather than a thread local, but changing that is a major break. Addressed that with `Kredis.global_namespace` in #158.

PR #158, opened 2024-12-19
Global namespacing that applies to all threads

jeremy:
Namespacing is thread-local currently, so it's easy to lose, e.g. when running parallel tests, Capybara tests, or executor pools.

Changing that is a major compatibility break, so we introduce a non-thread-local Kredis.global_namespace attribute that behaves as we'd expect.

When Kredis.namespace is set, it's appended to the global namespace.

### DIFF
--- a/lib/kredis/namespace.rb
+++ b/lib/kredis/namespace.rb
@@ -1,13 +1,30 @@
 # frozen_string_literal: true
 
 module Kredis::Namespace
-  def namespace=(namespace)
-    Thread.current[:kredis_namespace] = namespace
-  end
+  attr_accessor :global_namespace
 
   def namespace
-    Thread.current[:kredis_namespace]
+    if global_namespace
+      if value = thread_namespace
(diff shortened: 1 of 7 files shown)


maxim:
@jeremy FWIW, our parallelized test suite started failing randomly after upgrading to Rails 8 due to this issue. The solution in #156 didn't fix it for us (still seeing random failures). However, this `global-namespace` branch seems to completely eliminate the failures. Hoping to see it released soon.

ibrahima:
We also have been experiencing flaky tests recently, and it seems to be tests that depend on Kredis functionality. We also recently upgraded to Rails 8 so it could be the same issue. We do have the patch from #152 in our codebase so it seems like that's no longer enough to ensure that the namespace is set properly. It would be great to have this fix available in a released version so that our test suite can be more reliable. No pressure of course, we can try to work around it for now. I'll try out this branch to see if it helps. Thanks!

PR #163, opened 2025-06-01
Update README.md

alectrico-pro:
'value' instance method works like a getter method, it doesn't save input values.  Fixit by using  'default' setter method instead.


rafaelfranca:
What do you mean with it doesn't save input values? Implementation show it does https://github.com/rails/kredis/blob/main/lib/kredis/types/scalar.rb#L11

PR #164, opened 2025-06-01
Update README.md

alectrico-pro:
Ruby on Rails pluralize class names by convention. So Person model generate people when performing people=person.pluralize.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,6 +202,7 @@ end
 
 person = Person.find(5)
 person.names.append "David", "Heinemeier", "Hansson" # => RPUSH people:5:names "David" "Heinemeier" "Hansson"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -202,6 +202,7 @@ end
 
 person = Person.find(5)
 person.names.append "David", "Heinemeier", "Hansson" # => RPUSH people:5:names "David" "Heinemeier" "Hansson"
+Note: people is fabricate by "person".pluralize
```
rafaelfranca, line 205: I like the note but it is in the wrong place. We should for sure explain how the namespace is created, but it should be in free text, not in the code example, and it should also not show the `.pluralize` code.

